PR #4, opened 2015-02-23
Wrong backup command in Readme

thangngoc89:
Change from
``` bash
php artisan backup:start
```
to 
``` bash
php artisan backup:run
```


freekmurze:
Thanks!

PR #19, opened 2015-03-19
Restores the path property of FileSelector and update the test accordingly

MatthiasDeWinter:
Restores the path property of FileSelector and update the test accordingly.


freekmurze:
Thanks!

PR #20, opened 2015-03-19
Because ClassName::class available only in >= 5.5

2bj:
BackupCommand.php
line 70: $databaseBackupHandler = app()->make(DatabaseBackupHandler::class);

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         }
     ],
     "require": {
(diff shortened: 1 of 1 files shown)


freekmurze:
Hi, 

good catch, but instead of requiring a higher version of php, it's better to remove ::class and specify the full classname. So I'm not gonna merge this pull request. I'll change the ::class.

PR #26, opened 2015-04-15
Fix issue when no socket is defined

foxted:
A client of mine had some issues with this package recently: 

when exporting the database only without any socket defined, the exported file had only one table.


MatthiasDeWinter:
Hi,

Good spot!

I have tried your changes and it works like a charm, but when I run the tests they fail, even though Travis passes.

Can you try running the tests on your end? ( `$ vendor/bin/phpunit`)

foxted:
That's odd, the tests pass on my end as well!

freekmurze:
Thanks

PR #38, opened 2015-06-01
fix displaying different backup destination filename

ratiw:
By calling `getBackupDestinationFileName()` twice, the actual destination backup filename will be different than the one displaying when backup finish. Specifically in the 'second' digit.


freekmurze:
Thanks!

freekmurze:
Version 2.4.2 includes your fix.

PR #43, opened 2015-08-22
Add db only support

jkmassel:
Adds the ability to only backup the database, not files (as can be done from the command line)


freekmurze:
You can to this by specify an empty array in `source.files.include` in the config file.

PR #53, opened 2015-11-23
use Storage facade instead of alias

justutiz:
Laravel 5 don't have Storage alias by default.


freekmurze:
Laravel does have the Storage facade [since version 5.0.0](https://github.com/laravel/laravel/blob/v5.0.0/config/app.php#L191)

PR #56, opened 2015-12-14
fixed syntax error

JeroenVanOort:
How did this pass Travis? My PHP7 VM doesn't like it.


freekmurze:
Sorry about that. Thanks for your solution. Remember that requiring dev-master in your projects is like having unprotected sex with the internet :-)

PR #60, opened 2016-01-01
PostgreSQL support

Jamesking56:
Loving your library but I need PostgreSQL support.

Tried to keep this as tidy as possible, please feel free to edit. Its also missing compression support as your library seems to generate a .zip anyway (you can get `pg_dump` to generate a `.tar` file).

Hope this meets the coding standards, I did try my best!


freekmurze:
Thanks!

PR #62, opened 2016-01-07
Allow to write directly into an S3 bucket

### LINKED ISSUE #55: Using S3, not possible to write direct to bucket
I create S3 bucket dedicated for the backup and use IAM to provide permission to write to bucket.

Unfortunately it is not possible to set path='' to tell laravel-backup to write to the root of the bucket. Any slashes are removed and key is detected as being 0 characters long.

All works ok if folder is named but it mandates an extra folder level that otherwise is not required.

michellaurent:
When specifying a path in the config, dirname returns the name of the folder and makeDirectory ensures that this folder exists.

When the destination.path is empty, dirname returns a period "."

When makeDirectory is called with period as the directory name, AWS throws the exception.

A workaround is to skip the makeDirectory call when dirname returns a period.

Fixes #55.

### DIFF
--- a/src/Commands/BackupCommand.php
+++ b/src/Commands/BackupCommand.php
@@ -135,7 +135,9 @@ class BackupCommand extends Command
     {
         $destinationDirectory = dirname($destination);
 
-        $disk->makeDirectory($destinationDirectory);
+        if ($destinationDirectory != '.') {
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #69, opened 2016-02-07
Clarification of table headers

blueclock:
I think "number of backups" is clearer as a header because "amount of backups" could imply volume, not quantity.


freekmurze:
You're totally right. Thanks!

freekmurze:
A word to the wise: do not use this code in production (yet). It's not finished and still highly unstable :-)

PR #73, opened 2016-02-24
Add list command

xeno010:
Add new list command: `php artisan backup:list`

> List all backups

unbenannt

### DIFF
--- a/src/BackupServiceProvider.php
+++ b/src/BackupServiceProvider.php
@@ -40,7 +40,13 @@ class BackupServiceProvider extends ServiceProvider
             }
         );
 
-        $this->commands(['command.backup:run', 'command.backup:clean']);
+        $this->app['command.backup:list'] = $this->app->share(
+            function ($app) {
+                return new Commands\ListCommand();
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks for your contribution. Awesome feature. I'll review it in detail in the coming days.

freekmurze:
Some remarks on the output:
- the "create at" column should be renamed "Created at"
- it think it's better to output just one table for all backups. There should be a column named "Filesystem" that list the name of the filesystem the backup is saved on. 
- the "Expired in" column should be replace by and "Age" column that shows the age in days.

In your code you use several abbreviations like `$col` and `$v`. Please use full variable names, the are easier to read.

Could you make these changes? Then I'll pull.

xeno010:
Thank you very much for your feedback.

I have done all the points.

Here a new screenshot
unbenannt2

freekmurze:
Thanks!

freekmurze:
This code of this PR is included in version `2.10.0`.

PR #94, opened 2016-03-16
Fix typo

juukie:


freekmurze:
This is more than fixing a typo, it's fixing an actual bug.

Thanks!

juukie:
Indeed, I noticed I didn't get a failure notification :P

freekmurze:
I'll add some extra tests soon for the notification functionality so that this won't happen again.

Issue #59, opened 2015-12-26
 [Exception] Could not create backup of db


itzikbenh:
I installed the package on a new 5.1 project and I get this error when I run the artisan command. Can't figure out what is wrong.

PR #104, opened 2016-03-29
Added the ability to change the pushover notification sounds

tylercd100:
I think one of Pushovers best features is its different notification sounds.

With this change I made sure it wont break anybody's existing implementation if they don`t have the new config settings
```php
'sounds' => [
        'success' => env('PUSHOVER_SOUND_SUCCESS','pushover'),
        'error' => env('PUSHOVER_SOUND_ERROR','siren'),
],
```

`Line #59` @ `src/Notifications/Senders/Pushover.php` Makes sure they wont break

### DIFF
--- a/config/laravel-backup.php
+++ b/config/laravel-backup.php
@@ -161,8 +161,12 @@ return [
          * Here you can specify how messages should be sent to Pushover.
          */
         'pushover' => [
-            'token' => env('PUSHOVER_APP_TOKEN'),
-            'user'  => env('PUSHOVER_USER_KEY'),
+            'token'  => env('PUSHOVER_APP_TOKEN'),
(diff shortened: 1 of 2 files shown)


freekmurze:
Hi, thank you for putting some work into this!

The only thing I don't like is the hardcoded array with supported sounds.  I don't want it in the code because I don't want to maintain that array.  I assume pushover will use a default sound when using a unsupported one.

Could you please remove that array (together with the check if the sound is supported).

tylercd100:
Yes, Not a problem! I'll have an update shortly.

freekmurze:
I'll fix a few nitpicks of mine myself and tag a new release.

Thanks!!

PR #124, opened 2016-04-27
follow links

jorenvanhee:
FileSelection now follows symlinks. This is useful if you use a deployment tool like [Capistrano](http://capistranorb.com/documentation/getting-started/structure/).

### DIFF
--- a/config/laravel-backup.php
+++ b/config/laravel-backup.php
@@ -31,6 +31,11 @@ return [
                     base_path('node_modules'),
                     storage_path(),
                 ],
+
+                /*
+                 * Forces the following of symlinks if set to true.
+                 */
+                'followLinks' => false,
             ],
 
             /*
--- a/src/Tasks/Backup/BackupJobFactory.php
+++ b/src/Tasks/Backup/BackupJobFactory.php
@@ -32,7 +32,8 @@ class BackupJobFactory
     protected static function getFileSelection(array $sourceFiles)
     {
         return (new FileSelection($sourceFiles['include']))
-            ->excludeFilesFrom($sourceFiles['exclude']);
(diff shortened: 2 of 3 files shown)


freekmurze:
Thanks for your contribution. This is however a breaking change.

Let's make it non-breaking. Could you please introduce a config value `backup.source.files.followLinks` with a default value of `false`? Only if the config value is set to `true` the `followLinks` function may be called.

jorenvanhee:
I did it like this now

```php
$finder = (new Finder())
    ->ignoreDotFiles(false)
    ->ignoreVCS(false)
    ->files()
    ->in($directory);

if (config('laravel-backup.backup.source.files.followLinks')) {
    $finder->followLinks();
}
```

Everything works, but the tests fail ("ReflectionException: Class config does not exist").
I honestly don't have a lot of experience with testing, and don't know how to fix it.

freekmurze:
After reviewing this it don't like that `config` is called within the `FileSelection`-class. It this is better to add a `shouldFollowLinks`-method on the `FileSelection` class. 

You can read the config in [the `BackupJobFactory`-class](https://github.com/spatie/laravel-backup/blob/4969882d625dca8639514ff0c5aa610134412f25/src/Tasks/Backup/BackupJobFactory.php#L32) and call `shouldFollowLinks` there.

After you've done this modification the test 'll probably work again.

jorenvanhee:
The tests do work again, thanks for the hint!
Is this try better?

freekmurze:
Much better, thanks!

freekmurze:
These changes are available in the newly tagged version `3.7.0`.
Thanks again for your contribution.

PR #128, opened 2016-05-10
Added support for wildcards in the paths

stefro:
You can use a wildcard (*) in the include and exclude paths.
This is really useful in combination with https://github.com/spatie/laravel-medialibrary
Using a wildcard you're able to skip the backup of all the converted media. This will save a lot of backup storage in large projects.

example:
``` base_path('public/media/*/conversions') ```

### DIFF
--- a/config/laravel-backup.php
+++ b/config/laravel-backup.php
@@ -25,6 +25,10 @@ return [
                 /*
                  * These directories will be excluded from the backup.
                  * You can specify individual files as well.
+                 *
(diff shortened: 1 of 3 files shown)


freekmurze:
Awesome feature. Thank you very much

killgt:
Seems this featured has been deleted. I have the same use case, maybe would be usefull a config to choose a file selector class, so users can implement their own file selector.

PR #139, opened 2016-05-28
Better FileSelection

sebastiandedeyne:
Currently the `FileSelection` class lists all files that are excluded, and filters them out of the included file list. This implementation doesn't list all the excluded files, but just checks the paths with `starts_with`.

This should greatly improve speed since file IO is expensive, and in turn reduce memory usage.

Bonus: `FileSelection` has been refactored to pass around collections instead of arrays, cleaning up the code a bit.


freekmurze:
Looking good, great work! Will pull + do some tests + fix the unit test when I have a little bit of time on my hands!

PR #153, opened 2016-06-20
added support to --ignore-table mysqldump option

andremellow:
I've added an option to ignore some tables on MySql backups.

It can be done by adding an array with options to databases key
```
/*
+                * You can use some options to database backups
+                * includeTables - Incluide only these tables in backup
+                * excludeTables - Ingnore these tables in backup
+                * NOTE - You can't use includeTables and excludeTables at the same time.
+                * NOTE - Don't prefix the schema name. It'll be done automatically  
+                *'mysql' => [
+                *    'excludeTables' => [
+                *        'table_one' ,
+                *        'table_two'
+                *    ]
+                *]
+                * Currently only MySQL (includeTables and excludeTables) options are supported.
+                */
```

 in config file or to artisan command 

`php artisan backup:run --excluide-tables=table_1,table2`

I've also created some mysqldump tests. You must have a mysql server running to run the tests know

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 build
 composer.lock
 vendor
+phpunit.xml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to `laravel-backup` will be documented in this file.
 
+### 3.7.3 - 2016-06-21
+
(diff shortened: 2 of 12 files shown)


freekmurze:
I'll take care of this myself in the next major version of the package.

PR #161, opened 2016-07-06
Performance improvements

sebastiandedeyne:
Replaced file arrays with iterators to vastly reduce memory usage and speed up the backup if there are a lot of excluded files.


freekmurze:
flat 220x200 075 t u3

PR #177, opened 2016-07-27
fixes comment

jpmurray:
Comment for "allBackups" mentioned "all daily backups".

I kind of fixed those lines 5 minutes before realizing what was wrong ;-)


freekmurze:
Thanks!

PR #181, opened 2016-07-30
Add telegram notification

matriphe:
Add new notification method, using Telegram Bot API on events.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,8 @@
         "test": "vendor/bin/phpunit"
     },
     "suggest": {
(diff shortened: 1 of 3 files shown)


freekmurze:
Cool! Have you tested this with your own Telegram account?

matriphe:
yes, I. I use it for my projects. here's a screenshot of my Telegram notification.

photo_2016-08-07_13-42-49

if the chat_id is set for group chat, it could send the notification to a group.

freekmurze:
Thank you very much!

PR #189, opened 2016-08-15
Remove isReachable() check

### LINKED ISSUE #188: Write only mode for AWS
Is it possible to use a 'write only' mode, without listing/reading, for Amazon AWS?

Eg. I have a policy that only allows a 'PUT' command, so that a backup service can only ever put new backups and not read/delete them.

This combined with the Amazon versioning + lifecycle to move to glacier, makes sure that files are never deleted (unless the period is over). Do you have a recommended way to do this?

barryvdh:
Unnecessary listing of files as the write() method also throws an error when not reachable.
Fixes #188

### DIFF
--- a/src/BackupDestination/BackupDestination.php
+++ b/src/BackupDestination/BackupDestination.php
@@ -85,6 +85,10 @@ class BackupDestination
      */
     public function write($file)
     {
+        if (is_null($this->disk)) {
+            throw new Exception("Could not connect to disk {$this->diskName} because the disk is not set.");
+        }
+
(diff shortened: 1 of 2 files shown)


barryvdh:
Ehh, not really why half of the tests are failing though :/

freekmurze:
On first sight, me neither, but it will have to be figured out before I can pull this. I'll try to make some time for this soon.

barryvdh:
When the disk is null, a Fatal error is thrown, which is not caught. This was first checked in isReachable.

I added the check in the write() call, with a better error message.

freekmurze:
Perfect thanks!

PR #190, opened 2016-08-15
Make backup filename configurable

### LINKED ISSUE #188: Write only mode for AWS

barryvdh:
Allow set set the filename via `php artisan backup:run --filename=db.zip`
Fixes #188

### DIFF
--- a/src/Commands/BackupCommand.php
+++ b/src/Commands/BackupCommand.php
@@ -12,7 +12,7 @@ class BackupCommand extends BaseCommand
     /**
      * @var string
      */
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

PR #191, opened 2016-08-16
Cache composer, test on 7.1

barryvdh:
This will cache the composer downloads as zip, which makes the tests run faster.

Also added 7.1 to the matrix.


freekmurze:
Thank you, much appreciated!

PR #199, opened 2016-08-31
Allow for table(s) exclusion

antoniopaisfernandes:


freekmurze:
I'll add support for this myself in the next major version of the package.

antoniopaisfernandes:
Can you add exclusions in v3 branch please?

PR #250, opened 2016-09-13
Typo's and formatting

juukie:
I've only kept the styling of the events config because it looks better to keep it this way.


freekmurze:
Thanks!

PR #273, opened 2016-09-21
Use application name instead of url

vladimir-yuldashev:


freekmurze:
Going to leave it as it is for now. 

You can make this change in your own projects.

Issue #274, opened 2016-09-21
Exception thrown when trying to monitor backups


kfirba:
Hey.

I've just installed the package (V4) and tried playing with it for a bit.

# The Happy Path :smile:

<small>**Assumption: There are already few backups available**</small>

When I run `php artisan backup:monitor` I get the message that the backups are considered healthy/unhealthy and then an exception:

```
[Symfony\Component\Debug\Exception\FatalThrowableError]
  Call to undefined method Spatie\Backup\Tasks\Monitor\BackupDestinationStatus::getBackupDestination()
```

The fix is rather easy, seems like someone forgot to change the method call in `Spatie\Backup\Notifications\BaseNotification` on line 64:

```
// from:
return $this->event->backupDestinationStatus->getBackupDestination();

// to:
return $this->event->backupDestinationStatus->backupDestination();
```

However, unfortunately, this isn't all. Now when trying to run the monitor command I get the error:

```
[Symfony\Component\Debug\Exception\FatalThrowableError]
  Type error: Return value of Spatie\Backup\Notifications\Notifications\HealthyBackupWasFound::toMail() m
  ust be an instance of Illuminate\Notifications\Messages\SlackMessage, instance of Illuminate\Notificati
  ons\Messages\MailMessage returned
```

I then took a look at `Spatie\Backup\Notifications\Notifications\HealthyBackupWasFound` and found out that there is a mistake where someone who type-hinted the return type of the method `toMail` had type-hinted it to `SlackMessage` instead of `MailMessage`. To fix this we only need to fix the type-hinted return type:

```
// from:
public function toMail(): SlackMessage

// to:
public function toMail(): MailMessage
```

Now the errors are fixed and the notification will be triggered. I can see that the tests are WIP. Hopefully you add some tests to catch those tiny mistakes.

# The Sad Path :cry:

<small>**Assumption: There are no backups available**</small>

When trying to run the monitor command when there are no backups at all, the `UnhealthyBackupWasFound` event is fired. 

When the event is trying to build its problem description, it for some reason doesn't detect the fact that there are no backups yet. The reason is that for some reason someone negated the conditional on line 50:

```
// current:
if (! $backupStatus->amountOfBackups() === 0) {

// should be:
if ($backupStatus->amountOfBackups() === 0) {
```

The issue was that the above `if` statement returns a false positive just by accidentally adding the *!*. What happened then is it would fail on the condition on line 58:

```
if ($backupStatus->newestBackupIsTooOld()) {
[...]

PR #275, opened 2016-09-21
fix issues addressed in issue #274

kfirba:
Here is a pull request that adapts the corrections suggested in issue #274. 

I haven't added any tests. Let me know if you want me to add some tests.


freekmurze:
Thank you so much for this. I was planning on adding some tests for this functionality, but haven't found the time yet.

I'd appreciate a PR if you have the time.

Issue #163, opened 2016-07-11
Backup fails when backing up MySql db which has read / write hosts


drawmyattention:
I have a database connection defined as follows:

    'mysql' => [
            'read' => [
                'host' => env('DB_READ_HOST', 'localhost'),
            ],
            'write' => [
                'host' => env('DB_WRITE_HOST', 'localhost')
            ],
            'driver'    => 'mysql',
            // Other parameters
        ],

When I try to run ```php artisan backup:run```, I see the ```Backup failed because: Undefined index: host.``` error, because I am not setting a host directly.

There should be a check performed to check if there is a ```read``` host first, and if there isn't, use the ```host``` value of the database connection.

freekmurze:
Yup, this should be fixed. I'll try to make some time for this in the coming weeks. I'd also accept a PR if you need this quick.

bhulsman:
So let's say https://github.com/spatie/laravel-backup/blob/master/src/Tasks/Backup/BackupJobFactory.php could use some spice from https://github.com/illuminate/database/blob/master/Connectors/ConnectionFactory.php.

I'm I right?

freekmurze:
Support for read database connections has been added in `v4.1.0` of this package.

PR #293, opened 2016-10-19
Use read database connection if available

mpociot:
This PR fixes issue #163 
If a read database host is available, this will be used.
Otherwise, it will fall back to the default host key.


freekmurze:
Thank you!

Issue #277, opened 2016-09-27
change backup file name?


alinasiri1367:
How can I customize my backup file name?

freekmurze:
Right now, you cannot. But I'd accept a fully tested PR if the changes are relatively simple.

freekmurze:
This is now possible in the latest version of the package.

PR #295, opened 2016-10-20
Make backup filename configurable

mpociot:
This PR addresses issue #277 and allows the user to change the backup filename in the configuration.
I decided against putting the filename in the artisan command, because you have more control over it within the configuration file.

I needed to reset the config values in the test, because otherwise Carbon would return the previously used date.

### DIFF
--- a/config/laravel-backup.php
+++ b/config/laravel-backup.php
@@ -49,6 +49,11 @@ return [
         'destination' => [
 
             /*
+             * The filename prefix used for the backup zip file.
+             */
+            'filename_prefix' => '',
+
+            /*
              * The disk names on which the backups will be stored.
              */
             'disks' => [
--- a/src/Tasks/Backup/BackupJob.php
+++ b/src/Tasks/Backup/BackupJob.php
@@ -169,7 +169,7 @@ class BackupJob
     {
         consoleOutput()->info("Zipping {$manifest->count()} files...");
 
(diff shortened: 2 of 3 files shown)


config/laravel-backup.php
```diff
@@ -49,6 +49,11 @@
         'destination' => [
 
             /*
+             * The file name used for the backup zip file.
+             */
+            'filename' => \Carbon\Carbon::now()->format('Y-m-d-H-i-s'),
```
ninthspace: Doesn't this just set an invariant filename based on the current date and time, i.e. all backups would have the same filename? That's why the tests require manually overriding of the attribute.

Perhaps this should instead be a date format string which then gets instantiated for each backup.

freekmurze:
Thanks again Marcel for your work on this. 

I don't like the fact that users could now screw up the backup process by just using a fixed string as a backup name. Our package should make certain that the name of a backup is unique.

Could you change the `filename` key in the config file to `filename_prefix`? Default value is empty string. The package should then determine the filename like this:

prefix + `Carbon::now()->format('Y-m-d-H-i-s')` + '.zip'

Can you change your PR so it uses this behaviour?

mpociot:
Hey Freek,

just wanted to let you know that I've modified my PR and pushed the changes :)

freekmurze:
Thanks, I'll review (and probably pull) it soon!

freekmurze:
Perfect, thanks!

Issue #298, opened 2016-10-26
**[PROPOSAL]** Encrypting database backups


scottconnerly:
**In raising this issue, I confirm the following (please check boxes):**
- [X] I have read and understood the [contributors guide]().
- [X] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [X] I have checked that the bug-fix I am reporting can be replicated.
- [X] I have sent a postcard of my hometown.

---

I'm interested in securing the contents of my database backups by encrypting them. I've looked into multiple ways to do this and would like your input on which option(s) you would accept as a pull request:

1) Use password protected zip files
I'm aware that PHP's ZipArchive does not support creating encrypted zip files, and the only way to do such is to make a system call to the 'zip' command. This will require a bunch of if/else statements in \Spatie\Backup\Tasks\Backup\Zip, when no password is in play, use PHP's ZipArchive, when a password is in play, call the system 'zip' command. Since a key portion of the security of password protected zip files is the strength of the password, I would also add a minimum-length to the password used to create archives.

2) Encrypt the SQL file before including in manifest
It could either be explicit in BackupJob::dumpDatabases() or it could be an callback in createBackupManifest() that takes and returns (a possibly modified) $databaseDumps. While this is probably the cleanest from the point of view of this package, it isn't optimal for the file itself, because its better to compress then encrypt, rather than encrypt then compress. But it'd still get the job done. For my purposes, encryption is way more important than effective compression. However, this could allow for stronger encryption than Zip files offer.

3) Abstract \Spatie\Backup\Tasks\Backup\Zip so I can extend it and bind my class that I implement option 1 in. I would probably abstract BackupJob too in case I needed to extend that class too for currently unforseen reasons.

4) Allow BackupCollection to recognize non-zip files as well. I could trigger my encryption of already compiled zip files on the BackupWasSuccessful event. I'd want to change the file name from, for example, backup20161024.zip to backup20161024.zip.encrypted.

Do any of these approaches seem appealing to you?

freekmurze:
Hi,

thank you for describing your suggestions in such detail.

I'd like to keep this package as simple and generic as possible. Encrypting is something I don't want to maintain. 

The simplest thing to could be done to support your use case is to just manipulate the manifest. Currently files can only be added to the manifest. If it were to have some more functions such as search and remove it would be a lot more powerful. You could search the dumped version of the db, adding a encrypted copy of the file, and/or optionally removing the encrypted db dump.

[...]

PR #307, opened 2016-11-07
Encrypt backups

TheShiftExchange:
So from the discussion here https://github.com/spatie/laravel-backup/issues/298

I've made a basic PR for this, to show how simple it is. I've tested it, and it works perfectly. You can toggle encryption on/off in the config (default is off).

Encrypts the whole file prior to putting it in the various destination folders.

I feel this is a great addition to the package, so hopefully this (or a variation of) can be included...

### DIFF
--- a/config/laravel-backup.php
+++ b/config/laravel-backup.php
@@ -59,7 +59,7 @@ return [
         /*
          * Encrypt the final zip file using Laravel Crypt
          */
-        'encrypt' => false
+        'encrypt' => false,
     ],
 
 
--- a/config/laravel-backup.php
+++ b/config/laravel-backup.php
@@ -55,6 +55,11 @@ return [
                 'local',
             ],
         ],
+
+        /*
+         * Encrypt the final zip file using Laravel Crypt
+         */
+        'encrypt' => false
     ],
 
 
--- a/src/Tasks/Backup/BackupJob.php
+++ b/src/Tasks/Backup/BackupJob.php
@@ -4,6 +4,7 @@ namespace Spatie\Backup\Tasks\Backup;
 
 use Carbon\Carbon;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\Crypt;
 use Spatie\Backup\BackupDestination\BackupDestination;
 use Spatie\Backup\Events\BackupHasFailed;
 use Spatie\Backup\Events\BackupManifestWasCreated;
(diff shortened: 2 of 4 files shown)


freekmurze:
Hi,

thanks your for this cool PR. I didn't know the code change would be so simple.

Could you:
- add a test that verifies that this works? I need to be sure that when the new config value is `false` the backup is not encrypted. And when it is true it is encrypted. (for that last test you could just try to decrypt it)
- submit a PR to [the documentation](https://docs.spatie.be/laravel-backup/v4/installation-and-setup) updating the config file and explaining the future. There's a little `edit` button on the top of the page

TheShiftExchange:
@freekmurze - do you know if we can use `Crypt::shouldReceive('encrypt')->once();` in the unit tests?

It doesnt seem to be working - and it is getting beyond my package skills. Can you point me in the right direction to test the function based upon my PR?

freekmurze:
Hi,

after haven giving this some more though I've decided against merging this in. 

My main problem with this PR is that `file_get_contents` loads up the entire zip into memory, which could be problematic for large backups. Besides the provided Laravel function there are many other ways to perform encryption on file.

In version `4.2.0` I've added [a `BackupZipWasCreated`-event](https://docs.spatie.be/laravel-backup/v4/taking-backups/events#backupzipwascreated). You can listen for this event a perform the encryption you want on the path given in `$pathToZip`.

Though it'll not get merged in, I'd like to thank you for your work on this PR.

emohamed:
Just wanted to share a different approach to implement database dump encryption here(without actually requiring changes in the library).

In order to avoid loading the database file in memory, it is possible to pipe the dump command into `gpg` before saving it to the destination file. In order to do that, the `Compressor` option in `config/backup.php` would need to be setup to use something like this:

```
<?php
namespace App\DbEncryptor;
use Spatie\DbDumper\Compressors\Compressor as DbCompressor;

class GpgEncryptor implements DbCompressor {
    function useCommand(): string
    {
        return 'gpg --trust-model always -e -r ' . escapeshellarg(config('app.backup.encrypted_backups_recepient'));
    }

    function useExtension(): string
    {
        return 'gpg';
    }
}
```

`gpg` will compress the input while encrypting the stream, so it's not necessary to call `gzip` before or after the `gpg` call. 

In order for this approach to work, the system would need to have `gpg` installed. Additionally, the public key used for encryption would need to be imported in the default gpg keyring.

Also, the `app.backup.encrypted_backups_recepient` config value would need to be setup to point to the public key.

PR #308, opened 2016-11-11
Updated BaseNotification to return all values as string.

anchepiece:
Some slack-like tools (like mattermost) don't parse json as cleanly as slack does.


freekmurze:
Thanks

anchepiece:
Thank you! This package is easy to reason through. Its a narrow fix but fixes an issue I was dealing with.

Issue #314, opened 2016-11-22
Dump with routines on MYSQL


rafwell:
Hi, using mysql i need put `--routines` as parameter for `mysqldump`. I've tried using this on the mysql connection at `app/database.php` but not work.
`    'dump' => [
                'add_extra_option' => '--routines'
            ]
`

I've created a PR to solve issue: https://github.com/spatie/laravel-backup/pull/315

freekmurze:
v4 of the package can do this out of the box.

PR #315, opened 2016-11-22
Work with extraOptions

rafwell:
see https://github.com/spatie/laravel-backup/issues/314


freekmurze:
If you need some fine grained control over which options to pass to the db dumper, please upgrade to v4 of the package.

PR #324, opened 2016-12-10
Symlinks and Exclusions

### LINKED ISSUE #319: Exclude paths doesn't work in storage path with Envoyer
I have an issue with the backup of the files in the storage_path.
I use Envoyer for deployments, so the storage_path is outside the base_path (with symlinks).

Exclude rules doesn't work. All the files are in the zip ... 
On my local computer, the exclude rules works well... 

Here's my laravel-backup config file.

```
'files' => [

    /*
     */
    'include' => [
        storage_path(),
    ],

    /*
     *
     */
    'exclude' => [
        storage_path('backups'),
        storage_path('app/laravel-backup'),
        storage_path('framework'),
        storage_path('logs'),
    ],

    /*
     * Determines if symlinks should be followed.
     */
    'followLinks' => true,
],
```

If I use the default config file, the exclude rule works well...

Did I miss something ?

jpscharf:
- Fixed an issue with symlinked folders not honoring exclusions
Fixes #319

### DIFF
--- a/src/Tasks/Backup/FileSelection.php
+++ b/src/Tasks/Backup/FileSelection.php
@@ -114,7 +114,7 @@ class FileSelection
     protected function shouldExclude(string $path): bool
     {
         foreach ($this->excludeFilesAndDirectories as $excludedPath) {
-            if (starts_with($path, $excludedPath)) {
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #328, opened 2016-12-19
proper use of example.com domain instead of email.com

JonathanMH:
Hi, thank you so much for the awesome project! I just have a quick addition to not use a domain anybody could be listening to, but an official example domain, in case somebody forgets to change this.

Have a great week and a great christmas after that!


freekmurze:
Thanks!

Issue #340, opened 2017-01-04
Filename is not set


gtapps:
Hello guys,

I noticed the --filename option is not set when creating the zip file for a backup.
I'm going to create a pull request with the fix for this, which will update Tasks/Backup/BackupJob.

freekmurze:
If this is still and issue, please send a PR to solve it.

PR #341, opened 2017-01-04
#340 - Filename is not set when zipping.

gtapps:
Solves the problem with a default filename not being set.

### DIFF
--- a/src/Tasks/Backup/BackupJob.php
+++ b/src/Tasks/Backup/BackupJob.php
@@ -170,7 +170,7 @@ class BackupJob
     {
         consoleOutput()->info("Zipping {$manifest->count()} files...");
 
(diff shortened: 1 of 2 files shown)


freekmurze:
Great, could you also add a test to make sure that this does work?

d0minationPT:
Hello, sure, will take care of this today.

d0minationPT:
Done

freekmurze:
According to Travis your tests seem to contain an error. Could you fix that?

gtapps:
Seems like the issue with travis is because of files existing in the system?

freekmurze:
Your change broke existing tests. Your change doesn't take into account that setting a filename is optional. I'll fix this quickly myself.

Thank you for your work on this.

freekmurze:
Further investigation lead me to the root cause of the failing test. The default filename was set using php's `date` function which ignores the fixed datetime set by Carbon class in the tests.

gtapps:
Awesome, thank you for the good extensions!

PR #359, opened 2017-02-02
Changed BackupJob to only exclude the backup destination directory

### LINKED ISSUE #353: Backing up files in storage folder
One of my main goals with this package was to backup user-generated content in the storage/app folder (in addition to the database and code backup, which is great).  I noticed that this path seems to be automatically excluded - I'm thinking maybe because the backup is also in this path?  Is there a way to include specific folders under storage/app that you want included in the backup?  I have a short list, I've tried explicitly including it in the "include" config and also creating a a listener for BackupManifestWasCreated and adding them there, but neither method seemed to result in my files winding up in the zip file.

AlexVanderbist:
Fixes #353


freekmurze:
Great! Thank you very much!

Issue #358, opened 2017-02-01
Temporary directory doesn't exist, due to space char


graystevens:
It seems 4.4.2 introduced a bug with the temporary directory containing a space char:

`Backup failed because The dump process failed with exitcode 2 : Misuse of shell builtins : sh: 1: cannot create /tmp/0.83890900 1485990954/db-dumps/wisedb.sql: Directory nonexistent`

Rolling back to 4.4.1 fixes the issue.

marcusmoore:
Looks like [spatie-temporary-directory](https://github.com/spatie/temporary-directory) was added in 4.4.2. What operating system and does it happen consistently?

funcoding:
Even I am facing the same issue. I am using Ubuntu 14.04

freekmurze:
This problem has been fixed in the underlying `spatie/temporary-directory` package. 

Just run `composer update` to pull in the fix.

My apologies for any inconvenience.

dvlpp:
I'm following this thread because I'm experiencing the same issue, either on OSX and Ubuntu 14.04; I tried with the 4.4.4 version, after the https://github.com/spatie/laravel-backup/pull/360 PR, without success. For now, I downgraded to 4.4.1.

On OSX, error is:
`Backup failed because The dump process failed with exitcode 1 : General error : sh: /var/folders/zn/xqkqcz2x6z195fjlpq7gm2c00000gn/T/0.221896001486038964/db-dumps/radar-prod.sql: No such file or directory`

freekmurze:
We're working on a new fix.

PR #360, opened 2017-02-02
Fix version  constraints for package spatie/temporary-directory

RemiCollin:
Version constraints for `spatie/temporary-directory` was set to `~1.0.0`, which prevented composer from resolving the latest release of the package. Setting it to `~1.0` allows to depends on any `1.x`version. 

Fix  https://github.com/spatie/laravel-backup/issues/358


freekmurze:
Thanks!

PR #363, opened 2017-02-02
Changed temporary directory location and added some tests

AlexVanderbist:


mtx-z:
Hello,
is there any way to configure the zip temporary folder used ? 
I'm using a SSD, but it don't have enough space to contain a complete projet ZIP (lot of user uploaded files). 
I'd rather use a external HDD, to contain the ZIP, then upload to S3 from this disk. 

Thx !

PR #369, opened 2017-02-06
Trim starting `/` from filepath in backup archive to avoid empty root directory

### LINKED ISSUE #367: Nameless root folder in archive
Is there any reason the root folder in the backup zip would be nameless? Is this by design?

I opened the ZIP file on Windows and it showed an empty directory. I kept thinking something was broken. Later, I tried opening the file with 7zip which shows a nameless root folder that contains the db-dumps and home folders.

AlexVanderbist:
This should fix #367.


freekmurze:
Thank you!

Issue #378, opened 2017-02-15
Back up is ignoring excluded directories


bencarter78:
Hi

I'm currently using Laravel 5.4.11 with v3.5 of laravel-backup and I've noticed that when the backup runs it is ignoring the directories listed in the 'exclude' array, it's including all vendor and node_module sub-directories.

Any ideas why this might be? Is v3.5 compatible with L5.4? My server is running PHP 5.6 at the moment so upgrading to v4 isn't an option at the minute.

Thanks

AlexVanderbist:
Hi Ben,

Is there are reason you are using `v3.5` instead of the latest release for v3 (`v3.10.2`)? If not, could you try updating to `3.10.2` and let me know if that solves the issues?

Thanks!

bencarter78:
Hi @AlexVanderbist I have updated to 3.10.1 (3.10.2 requires L5.1|5.2|5.3) but it's still ignoring the exclude array

bencarter78:
I think I've worked out the problem, it fails the shouldExclude method because in Laravel 5.4 the contains method has had it's signature altered ($key and $value have swapped).

Basically, it seems like v3.* is only compatible with Laravel 5.3 and below.

AlexVanderbist:
Hi, I just stumbled uppon the same block of code. The key/value order change has already been introduced in Laravel 5.3 tho, that's why v3.10.2 of this package removed `contains` from the `shouldExclude` method altogether.

This means the package _might_ just work for L5.4. Feel free to fork the package, bump the required version in `composer.json` and test this out.

PR #384, opened 2017-02-17
Updated composer.json v3 to be compatible Laravel 5.4 as well

dersvenhesse:
I noticed an issue regarding the `exclude` array while running Laravel 5.4 with PHP < 7. The folders listed in the `exclude` array aren't excluded. This results in massively increasing backup files. The problem is described in #378 as well.

So I updated the `composer.json` in v3 to be compatible with Laravel 5.4 too.
My backups worked fine afterwards. The `exclude` folders are excluded as desired. ;)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,9 +19,9 @@
     ],
     "require": {
         "php" : "^5.5|^7.0",
-        "illuminate/support": "~5.1.0|~5.2.0|~5.3.0",
-        "illuminate/console": "~5.1.0|~5.2.0|~5.3.0",
-        "illuminate/filesystem": "~5.1.20|~5.2.0|~5.3.0",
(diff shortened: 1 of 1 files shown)


freekmurze:
With this update the problem is resolved for users of L5.4, but not for users of L5.3. Could you rewrite `shouldExclude` in `FileSelection` so it doesn't use the `contain` method. If we use regular php functions there the problem will be fixed for both users of L5.3 en L5.4

dersvenhesse:
Version 3.10.2 is already without `contains`.

freekmurze:
Thanks!

freekmurze:
I've tagged a new release `3.11.0` that includes your changes.

PR #386, opened 2017-02-18
Using badge-poser

dersvenhesse:
Using badge-poser and packagist to show the latest stable version instead of the latest release in `README.md`. So the v3 releases won't be in the badge event if they are newer than a v4 release.


freekmurze:
Great, thanks! I'm probably going to use badge-poser for all our future package

PR #387, opened 2017-02-18
Added backupName to backupDestinationProperties

kapersoft:
Adding backupName to backupDestinationProperties makes the overview more clear when monitoring multiple backups.


freekmurze:
Thanks!

PR #395, opened 2017-03-14
Check for empty string in notifications config

### LINKED ISSUE #355: Emails are still send when the notification is set as an empty string
My config file is as follows:

```
    'notifications' => [
        'notifications' => [
            \Spatie\Backup\Notifications\Notifications\BackupHasFailed::class         => ['mail'],
            \Spatie\Backup\Notifications\Notifications\UnhealthyBackupWasFound::class => ['mail'],
            \Spatie\Backup\Notifications\Notifications\CleanupHasFailed::class        => ['mail'],
            \Spatie\Backup\Notifications\Notifications\BackupWasSuccessful::class     => [''],
            \Spatie\Backup\Notifications\Notifications\HealthyBackupWasFound::class   => [''],
            \Spatie\Backup\Notifications\Notifications\CleanupWasSuccessful::class    => [''],
        ],
        /*
         * Here you can specify the notifiable to which the notifications should be sent. The default
         * notifiable will use the variables specified in this config file.
         */
        'notifiable' => \Spatie\Backup\Notifications\Notifiable::class,
```

I am still receiving emails for all events, not just those where the array is not an empty string. I'd say this is not expected behaviour. However, I cannot seem to find the part of the source of this package where this happens so I can debug it.

It is important to note that when setting the array as an empty array `[]` emails are indeed not sent. So it's a minor thing, but I don't think it should happen.

AlexVanderbist:
If an array with an empty string `[""]` or simply an empty string `""` is used as the notification channel we'll default it to an empty array `[]` so Laravel doesn't get confused and starts sending mails.

This should fix #355

### DIFF
--- a/src/Notifications/BaseNotification.php
+++ b/src/Notifications/BaseNotification.php
@@ -15,7 +15,13 @@ abstract class BaseNotification extends Notification
      */
     public function via()
     {
-        return config('laravel-backup.notifications.notifications.'.static::class);
+        $notificationChannels = config('laravel-backup.notifications.notifications.'.static::class);
+
+        if (empty($notificationChannels) || empty($notificationChannels[0])) {
+            return [];
+        }
+
+        return $notificationChannels;
     }
 
     public function applicationName(): string


freekmurze:
Thanks!

PR #403, opened 2017-03-27
fix there is an errro when backup with some adapter fclose the resource internally

billyct:
```
Copying zip failed because: fclose(): supplied resource is not a valid stream resource.
```
with this error it also can work fine 😊


freekmurze:
Hi, thank you for this, meanwhile it has been fixed by another PR.

Issue #404, opened 2017-03-31
Unclear readme / installation instructions


bkuhl:
The readme for this project provides installation instructions on the front page.  However, they aren't for the latest version of Laravel.  Advanced users will usually jump right to installation instructions for a package they're familiar with what it does.  I think this should be removed and the location of the original installation instructions made more clear.

PR #405, opened 2017-03-31
Clarity on installation instructions

bkuhl:
Attempts to address https://github.com/spatie/laravel-backup/issues/404

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,21 +32,9 @@ The best postcards will get published on the open source page on our website.
 
 ## Installation and usage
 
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you for this. I've manually added a sentence to make the instructions a bit more clear.

bkuhl:
I appreciate the prompt feedback, but I don't think more words is the solution. Visually when I look at the read me the first thing I see is how to composer install an outdated version rather than the most recent.  That is because it is wrapped in a code block which causes it to stand out over all the text on the screen.

PR #406, opened 2017-03-31
Only close resource if not already done by Flysystem. 

JeroenJochems:
Flysystem adapter should do this. Made closing conditional so it also works if an implementation doesn't.


freekmurze:
Thanks!

PR #414, opened 2017-04-10
Adding ability to change the slack channel

lukzgois:
Add the ability to change the channel of slack notifications.


freekmurze:
Thank you for this PR. What happens if the channel is set to `null`?

lukzgois:
The default channel of the slack webhook is used.

freekmurze:
Thanks!

PR #419, opened 2017-04-14
Ukrainian translation

MammutAlex:


freekmurze:
Thank you, the iso code for Ukrainia seems to be `UA` is that correct? If so can you update your directory name?

MammutAlex:
Iso code for country (ISO 3166-1) is `UA ` iso code language (ISO 639-1) is `UK`

freekmurze:
Ah, ok, thanks for clarifying that.

PR #422, opened 2017-04-19
Optimise backup:list for external file systems

AlexVanderbist:
I noticed running `backup:list` with 2 backups on a disk generated about 90 filesystem API requests.

This PR caches the `BackupCollection` and the `BackupCollection` size when running the `backup:list` command and brings that number down to 20 API requests.

screen shot 2017-04-19 at 15 27 11

(still not ideal but this could be further improved by using [memory caching](https://flysystem.thephpleague.com/caching/) when the adapter is registered in the application)

### DIFF
--- a/src/BackupDestination/BackupDestination.php
+++ b/src/BackupDestination/BackupDestination.php
@@ -22,6 +22,9 @@ class BackupDestination
     /** @var Exception */
     public $connectionError;
 
+    /** @var null|\Spatie\Backup\BackupDestination\BackupCollection */
+    protected $backupCollection = null;
+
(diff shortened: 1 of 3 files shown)


src/BackupDestination/BackupDestination.php
```diff
@@ -93,9 +96,13 @@ public function backupName(): string
 
     public function backups(): BackupCollection
     {
-        $files = $this->isReachable() ? $this->disk->allFiles($this->backupName) : [];
+        if ($this->backupCollection) {
```
freekmurze: Rename this to `backupCollectionCache`

Issue #372, opened 2017-02-06
Extra option for using gzip


afraca:
Hi, 

I have some large databases. When we do manual backup with PhpMyAdmin or Adminer, it's nice to see that when we tick the gzip box, we often get impressive compression (yes, that depends on the data, irrelevant). 

It would be great if there is some extra option we can pass to get the same with the backups we create with this package. Any thoughts?

freekmurze:
By default this package will store the db dump in a zip file.

afraca:
Yes, I know... The DB dumps from PMA and Adminer can be 
.sql
or
.sql.gz 

The latter are a lot more compact. The default zip from laravel-backup does NOT do compression as far as I know, it's simply bundling files.

freekmurze:
I'd accept a PR that adds compression to the backups.

afraca:
Originally I opened the issue in spatie/db-dumper because I think it's easier to redirect the sql dump file through gzip than to change how the zipping works in **this** package. For my particular use-case _that_ would still be viable work, and a lot of gain.

Changing how **this** package deals with zips, currently with the built-in ZipArchive is trickier I suspect. The easiest would be something like 
* Create the .zip normal
* Pipe the .zip through gzip to create a new .zip.gz file or something

It's not standard at all to do that. That's why 
https://github.com/spatie/db-dumper/issues/35

Is still the preferred place to begin with this, IMHO.

freekmurze:
We'll handle this further in #424

PR #424, opened 2017-04-21
Add feature to optionally gzip dumped .sql

pdewit:
According to issue #372, add an option to gzip the SQL files before zipping them. Very useful in my use case since .sql.gz files are a lot smaller while still able to be loaded into my SQL client, unlike .zip files.

### DIFF
--- a/config/laravel-backup.php
+++ b/config/laravel-backup.php
@@ -46,6 +46,11 @@ return [
             ],
         ],
 
+        /*
+         * Gzip the dumped SQL files for extra compression
+         */
+        'gzipSql' => false,
+
         'destination' => [
 
             /*
--- a/config/laravel-backup.php
+++ b/config/laravel-backup.php
@@ -49,7 +49,7 @@ return [
         /*
          * Gzip the dumped SQL files for extra compression
          */
-        'gzipSql' => false,
+        'gzip_sql' => false,
 
         'destination' => [
 
--- a/src/Tasks/Backup/BackupJob.php
+++ b/src/Tasks/Backup/BackupJob.php
@@ -4,8 +4,8 @@ namespace Spatie\Backup\Tasks\Backup;
 
 use Exception;
 use Carbon\Carbon;
-use Spatie\Backup\Helpers\Format;
 use Spatie\DbDumper\DbDumper;
+use Spatie\Backup\Helpers\Format;
 use Illuminate\Support\Collection;
 use Spatie\Backup\Events\BackupHasFailed;
 use Spatie\Backup\Events\BackupWasSuccessful;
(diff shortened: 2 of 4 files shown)


config/laravel-backup.php
```diff
@@ -46,6 +46,11 @@
             ],
         ],
 
+        /*
+         * Gzip the dumped SQL files for extra compression
+         */
+        'gzipSql' => false,
```
freekmurze: use snake case here

src/Tasks/Backup/BackupJob.php
```diff
@@ -202,10 +203,48 @@ protected function dumpDatabases(): array
...
+
             return $temporaryFilePath;
         })->toArray();
     }
 
+    /**
+     * @param $temporaryFilePath
+     * @return string
+     */
+    protected function gzipSqlFile($dbDumper, $temporaryFilePath)
```
freekmurze: Refactor this to it's own class that is responsible for gzipping

pdewit:
Thanks for the feedback! I hope it's better now.

src/Tasks/Backup/Gzip.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        $gzipPath = $this->originalFilePath.'.gz';
+
+        if ($gzipOut = gzopen($gzipPath, 'w9')) {
+            if ($gzipIn = fopen($this->originalFilePath, 'rb')) {
+                while (! feof($gzipIn)) {
+                    gzwrite($gzipOut, fread($gzipIn, 1024 * 512));
+                }
+                fclose($gzipIn);
+            } else {
```
freekmurze: I should have also mentioned that I really dislike `else` statements. Could you refactor this code to avoid them entirely?

pdewit:
Yeah I can see that, the error handling also didn't work correctly that way so I added a test for that and fixes it. Hope it's better now 😃

freekmurze:
Cool, thanks

Holding of on this for a while, because we might use gzip for compressing the other files as well. When I have some time on my hands I'll think this through.

afraca:
Hey @pdewit  , thanks for doing this. As we only use the database backup feature, it's fine as it is, but I understand the plan from @freekmurze . 

(I know if I really want it, I should do it myself. Thanks for all your hard work guys) 
If you happen to approach this yourselves, could you make sure the generated database backup file can be imported by PMA / Adminer? They support gzipped files.

pdewit:
@afraca you can use my fork in the meantime, I'm not sure if it's going to be kept up to date but I will use it in production myself. https://github.com/pdewit/laravel-backup/tree/gzip-sql

freekmurze:
I've decided to merge this in so you can already make use of this.

In the next major version (no ETA on that yet, probably early 2018) I plan to make this kind of compression a first class citizen in the package.

PR #426, opened 2017-05-01
Call to undefined method getFilesystemName()

drbyte:
Call to undefined method Spatie\Backup\Tasks\Monitor\BackupDestinationStatus::getFilesystemName()


freekmurze:
Thanks!

PR #440, opened 2017-05-13
use different suffix depending on database system for file being dumped

vesper8:
before it would always default to .sql which caused a loss of backup in the case of both a mongodb and mysql database being dumped with the same database name

now it will dump both correctly


freekmurze:
Thank you for this. Although you changes will work, I prefer the way #441 goes about solving this problem.

vesper8:
all good

PR #441, opened 2017-05-13
Updated database dump file name generation to prevent overwrite.

### LINKED ISSUE #439: mongodb backups have the .sql suffix in filename, leading to loss of backup if backing up dbs with same name
I am backing up both a mysql and mongodb database. As it happens, the name of the database is the same for both database systems "mydb". And I have gzip enabled

So first it backed up the mysql and gzipped it to mydb.sql.gz, then the mongodb was backed up and the name was the same mydb.sql.gz (should be mydb.mongo.gz)

then it was packaged into a file with datestamp, but by then the sql zipped had been overwritten

SandeeshS:
This fixes #439

Appends the database type to the file name to prevent multiple databases with the same name from being overwritten during backup.

### DIFF
--- a/src/Tasks/Backup/BackupJob.php
+++ b/src/Tasks/Backup/BackupJob.php
@@ -6,6 +6,7 @@ use Exception;
 use Carbon\Carbon;
 use Spatie\DbDumper\DbDumper;
 use Illuminate\Support\Collection;
+use Spatie\DbDumper\Databases\Sqlite;
 use Spatie\Backup\Events\BackupHasFailed;
 use Spatie\Backup\Events\BackupWasSuccessful;
 use Spatie\Backup\Events\BackupZipWasCreated;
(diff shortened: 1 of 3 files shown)


freekmurze:
Could you add a test to make sure that this works?

SandeeshS:
Sure i'll add a test and update the PR

SandeeshS:
@freekmurze I just found an issue with the **db-dumper** package when trying to setup a test with sqlite database. Has the package ever been tested with a real sqlite db after adding support? Currently dumping a sqlite database throws the following exception.

`Spatie\DbDumper\Exceptions\DumpFailed: The dump process failed with exitcode 1 : General error : Error: near line 1: near "$BEGIN": syntax error`

I'll look into it and try to open an issue and possibly fix in that package.

SandeeshS:
Few things to note.

- The old code was using the complete path of sqlite databases as the file name of the backup.
- The db-dumper package fails in dumping sqlite databases.

I've made it so that sqlite database names are set to '**database**' instead of the path when dumping. Therefore the final backup file would be **sqlite-database.sql**

Created a PR https://github.com/spatie/db-dumper/pull/49 to fix the db-dumper package. Merge the db-dumper fix before merging this PR for passing test.

tests/TestModel.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace Spatie\Backup\Test;
+
+use Illuminate\Database\Eloquent\Model;
+
+class TestModel extends Model
```
freekmurze: What is the use of this model? Can we remove it?

SandeeshS: This model is used to add sample data to the temporary sqlite db. This can be removed if needed since the backup dump would contain at least the table creating queries even without sample rows.

SandeeshS: I removed it now for simplicity sake.

freekmurze:
Great! Thanks!

phadaphunk:
@freekmurze This works fine but won't work for same type of databases with the same name on different machines / environments. Would you accept a PR that appends the host name as well given a certain parameter in the command?

PR #454, opened 2017-06-02
Support Auto-Discovery

rohithmeethal:


freekmurze:
Thanks you for this.

Note however that we'll only tag a L5.5 compatible version of the package when the L5.5 is released.

Issue #466, opened 2017-06-14
Unable to use Database notifications


jove4015:
I struggled to implement database notifications with this class.  The key issues were:

a) The Notifiable class did not extend Model, which the DatabaseNotification class was expecting

b) There was no toArray or toDatabase method on the notifications

I'm working on a pull request now.  It may not be perfect but at least it allows it to work.  I also had to add a migration to allow for a NULL notifiable_id in the notifications table, since the Notifiable "Model" does not have any ID per se in this case.

Altogether, this feels clunky.  I think it may be worth updating the database notifications themselves to make such a relationship optional - they seem to assume that all notifications will go to Users of the application, which isn't really practical in the context of CLI command.

freekmurze:
We'll handle this further in #467

PR #467, opened 2017-06-14
allow database notifications to work

jove4015:
See issue #466

### DIFF
--- a/src/Notifications/BaseNotification.php
+++ b/src/Notifications/BaseNotification.php
@@ -34,6 +34,11 @@ abstract class BaseNotification extends Notification
     {
         return $this->backupDestination()->diskName();
     }
+    
+
+    public function toArray($notifiable) {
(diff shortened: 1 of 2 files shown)


src/Notifications/BaseNotification.php
```diff
@@ -35,6 +35,11 @@ public function diskName(): string
         return $this->backupDestination()->diskName();
     }
 
+    public function toArray($notifiable)
+    {
+        return $this->backupDestinationProperties();
```
freekmurze: Could you add a test to prove that this works?

freekmurze:
I'll pull this in after you've add a test to prove that it works.

freekmurze:
Could you add a test?

freekmurze:
Closing due to inactivity. Reopen if you want to work a bit more on this

Issue #500, opened 2017-08-11
FileSelectionTest fails in linux ubuntu


jago86:
Hi, I'm running the tests in my local environment: Ubuntu 16.04, PHP 7.0.2 and i get some failures like this:

```
1) Spatie\Backup\Test\Unit\FileSelectionTest::it_can_exclude_files_from_a_given_subdirectory
Failed asserting that Array &0 (
    0 => '/home/jairo/proyectos/laravel-backup/tests/stubs/file1.txt'
    1 => '/home/jairo/proyectos/laravel-backup/tests/stubs/.dotfile'
    2 => '/home/jairo/proyectos/laravel-backup/tests/stubs/directory2/directory1/file1.txt'
    3 => '/home/jairo/proyectos/laravel-backup/tests/stubs/file2.txt'
    4 => '/home/jairo/proyectos/laravel-backup/tests/stubs/file3.txt'
) is identical to Array &0 (
    0 => '/home/jairo/proyectos/laravel-backup/tests/stubs/.dotfile'
    1 => '/home/jairo/proyectos/laravel-backup/tests/stubs/directory2/directory1/file1.txt'
    2 => '/home/jairo/proyectos/laravel-backup/tests/stubs/file1.txt'
    3 => '/home/jairo/proyectos/laravel-backup/tests/stubs/file2.txt'
    4 => '/home/jairo/proyectos/laravel-backup/tests/stubs/file3.txt'
).
```
Apparently, the FileSelection class is getting the correct files but in a different order than the first array, so I think it is false positive. I can make some changes in the test to fix this if actually is a bug in the test.

freekmurze:
Strange that the tests on Travis (that use PHP 7.0) indicate no errors. 

But sure, I'd accept a easy fix for this.

PR #502, opened 2017-08-14
Fix test failure

jago86:
Selection of files ($fileSelection->selectedFiles())  seems that in some environments generates an unordered array giving a false positive #500 . Sort the arrays before compare it fixes the problem.

### DIFF
--- a/tests/Unit/FileSelectionTest.php
+++ b/tests/Unit/FileSelectionTest.php
@@ -22,8 +22,7 @@ class FileSelectionTest extends \PHPUnit_Framework_TestCase
     {
         $fileSelection = new FileSelection($this->sourceDirectory);
 
-        $this->assertSame(
-            $this->getTestFiles([
+        $testFiles = $this->getTestFiles([
                 '.dotfile',
                 '1Mb.file',
                 'directory1/directory1/file1.txt',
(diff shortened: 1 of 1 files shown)


tests/Unit/FileSelectionTest.php
```diff
@@ -34,9 +33,13 @@ public function it_can_select_all_the_files_in_a_directory_and_subdirectories()
...
-            ]),
-            iterator_to_array($fileSelection->selectedFiles())
-        );
+            ]);
+        $selectedFiles = iterator_to_array($fileSelection->selectedFiles());
+
+        sort($testFiles);
+        sort($selectedFiles);
+
+        $this->assertSame($testFiles, $selectedFiles);
```
freekmurze: Instead of doing this check over and over again could you add a new protected method `assertSameArray` where in you sort both arrays and perform `assertSame`?

jago86:
@freekmurze, sounds good! I just push the changes requested.

tests/Unit/FileSelectionTest.php
jago86: @freekmurze, sounds good! I just push the changes requested.

tests/Unit/FileSelectionTest.php
```diff
@@ -34,9 +33,10 @@ public function it_can_select_all_the_files_in_a_directory_and_subdirectories()
                 'file1.txt',
                 'file2.txt',
                 'file3.txt',
-            ]),
-            iterator_to_array($fileSelection->selectedFiles())
-        );
+            ]);
+        $selectedFiles = iterator_to_array($fileSelection->selectedFiles());
```
freekmurze: Brrr, is this ugly `iterator_to_array` thing really needed? 😄

jago86: Well, that was there since the first time that I clone the repo, and it seems is necessary because the method returns an Generator object.

freekmurze:
Thanks!

PR #504, opened 2017-08-21
Adding rolling strategy.

websanova:
Just though I would drop this in, in case someone else would like to use.


freekmurze:
Thanks for this, but I'll pass. I'm only going to maintain the current default strategy.

websanova:
It's 5 lines of code but ok.

PR #508, opened 2017-08-28
Empty the Temp directory before starting a backup job

willoller:
There is an edge case where the temp file fills up until it is using all available space, causing future backups to fail. Since a failed backup does not clean the temp directory, the backup process will never succeed without intervention.

This is solved by always clearing the temp file before starting the backup job.


freekmurze:
Thanks!

PR #525, opened 2017-09-22
[WIP] Makes backups to be created at exact specified location locally and fixes #185

### LINKED ISSUE #185: Local Backup Destination Problem
Hi,

I'm using version 3.9.0.

And i noticed that backup system creates a folder storage >> laravel-backup But backup folder is empty. I searched backup name and i found files in storage >> app >> http---localhost

How can we solve this problem ?

Thank you.

introwit:
- Resolves #185 
- The `temp` directory is now created in the specified backup folder itself and removes creation of an empty `backup` folder in `/storage/app` directory.
- `APP_NAME` is used instead of `APP_URL` for the name property which I think would be more appropriate.
- Removed unused imports.

Open to suggestions and feedback :)

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -8,7 +8,7 @@ return [
          * The name of this application. You can use this name to monitor
          * the backups.
          */
-        'name' => env('APP_URL'),
+        'name' => env('APP_NAME'),
 
         'source' => [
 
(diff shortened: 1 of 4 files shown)


freekmurze:
Tests seem to fail, could you try to fix them?

freekmurze:
I've solved this issue by simply renaming the temporary folder that gets created to `backup-temp`.

PR #533, opened 2017-09-29
Use APP_NAME instead of APP_URL

introwit:


phpunit.xml.dist
```diff
@@ -20,6 +20,7 @@
         </whitelist>
     </filter>
     <php>
+        <env name="APP_NAME" value="mysite"/>
         <env name="APP_URL" value="mysite.com"/>
```
freekmurze: Can this be removed?

freekmurze:
Nice, thanks!

PR #536, opened 2017-09-30
fix CleanupHasFailed application_name translation not working

chrisrenga:
fix the :application_name not getting translated for failed backup cleanup notifications


freekmurze:
Thanks!

PR #539, opened 2017-10-04
correct link to v4 in README.md

MikkyX:
I noticed the link to v4 in the "Using an older version..." section was actually pointed to v3, so I took a moment to correct it


freekmurze:
Thanks!

PR #562, opened 2017-11-03
Fix deleting all backups when using maximum storage

### LINKED ISSUE #505: backup: clean to s3/local removing all files but one w/ error
Hello

Running backup:clean for the following config will result in an error

> Cleanup failed because: File not found at path: backups/site_2017-08-21-16-07-13.zip

For config

```
'cleanup' => [

        'strategy' => \Spatie\Backup\Tasks\Cleanup\Strategies\DefaultStrategy::class,

        'defaultStrategy' => [

            /*
             * The number of days for which backups must be kept.
             */
            'keepAllBackupsForDays' => 7,

            /*
             * The number of days for which daily backups must be kept.
             */
            'keepDailyBackupsForDays' => 16,

            /*
             * The number of weeks for which one weekly backup must be kept.
             */
            'keepWeeklyBackupsForWeeks' => 8,

            /*
             * The number of months for which one monthly backup must be kept.
             */
            'keepMonthlyBackupsForMonths' => 4,

            /*
             * The number of years for which one yearly backup must be kept.
             */
            'keepYearlyBackupsForYears' => 2,

            /*
             * After cleaning up the backups remove the oldest backup until
             * this amount of megabytes has been reached.
             */
            'deleteOldestBackupsWhenUsingMoreMegabytesThan' => 6000,
        ],
```

Whats actually happening is it removes that file that says it's missing (in fact, all the files except the newest one) and then tries to delete it again, and results an error.

I feel like the date created is not being properly used, it happens with both s3 and local storage

Will try to dd and figure out whats the problem with the date stamps

AlexVanderbist:
This PR fixes a bug in `DefaultStrategy` where all backups would be removed when cleaning old backups due to maximum storage being reached (see #505).

This PR also fixes the `it_can_remove_old_backups_until_using_less_than_maximum_storage` test to catch the issue if it were to happen again.

Finally the testing `APP_NAME` environment variable is set to `mysite` in `phpunit.xml.dist` however in a lot of the tests `mysite.com` was used. I've made all the tests use the same `APP_NAME` now for consistency.

Fixes #505

### DIFF
--- a/src/Tasks/Cleanup/Strategies/DefaultStrategy.php
+++ b/src/Tasks/Cleanup/Strategies/DefaultStrategy.php
@@ -115,6 +115,8 @@ class DefaultStrategy extends CleanupStrategy
 
         $oldest->delete();
 
+        $backups = $backups->filter->exists();
+
         $this->removeOldBackupsUntilUsingLessThanMaximumStorage($backups);
     }
 }
--- a/tests/Integration/CleanupCommandTest.php
+++ b/tests/Integration/CleanupCommandTest.php
@@ -27,7 +27,6 @@ class CleanupCommandTest extends TestCase
     /** @test */
     public function it_can_remove_old_backups_until_using_less_than_maximum_storage()
     {
-        // 1 megabyte storage size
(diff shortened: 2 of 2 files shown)


freekmurze:
Thanks!

PR #563, opened 2017-11-04
Feature disable backups per environment

edgarnadal:
I have an application installed on several environments and some needs backup and some other no, so I need to have a way to disable it by configuration.

This PR adds the ability to disable the backups depending on a config. By default all backups will be enabled, just need to add `BACKUP=false` to the _.env_ file.

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -5,6 +5,11 @@ return [
     'backup' => [
 
         /*
+         * Flag to determine if the backup should be run on the current env
+         */
+        'run' => env('BACKUP', true),
+
+        /*
          * The name of this application. You can use this name to monitor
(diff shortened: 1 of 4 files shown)


freekmurze:
Thanks for your work on this but I'm not going to pull this is.

You can easily use the scheduler to schedule things in a specific environment:

```php
// app/Console/Kernel.php

    protected function schedule(Schedule $schedule)
    {
       if (app()->environment('production')) {
           $schedule->command('backup:run')->daily();
       }
    }
```

edgarnadal:
I understand but sometimes you have client A that needs backup and client B don’t.

But thanks anyway.

PR #580, opened 2017-11-25
Fixing styleci + using tap method to refactor some variable declarations

arcanedev-maroc:


freekmurze:
Going to leave things as they are for now.

PR #584, opened 2017-11-26
Use assertFileExists and assertFileNotExists

gabriel-caruso:
Small one: use [`assertFileExists`](https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertFileExists) and [`assertFileNotExists`](https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertFileExists).


freekmurze:
Better indeed, thanks!

PR #589, opened 2017-12-15
Append env to app name in slack notification

RoobieBoobieee:


freekmurze:
Going to leave this as is for now.

PR #593, opened 2018-01-06
Fixes from PHPStan

carusogabriel:
After running [`PHPStan`](https://github.com/phpstan/phpstan) up to level 2, some code could be refactored :smile:


freekmurze:
Thanks!

Issue #594, opened 2018-01-06
[Feature] add an event b4 the backup start


ctf0:
atm b4 the package start to backup the db it will lock it to prevent any new data being inserted, but if the user at the time of the backing up tried to add a new model he would get an error because the db is currently locked.

so i was thinking maybe we can have an extra event that fires just b4 the db gets locked, this way we can listen to it and show like an alert/notification to the user while the backup is being created & later we can use `BackupZipWasCreated` to hide that alert.

freekmurze:
I'd accept a PR with tests that adds a `BackupStarted` event.

PR #596, opened 2018-01-07
add BackupStarted event

ctf0:
u can use it like
```php
Event::listen('Spatie\Backup\Events\BackupStarted', function () {
    return logger(‘db backup has started’);
});
```

https://github.com/spatie/laravel-backup/issues/594

### DIFF
--- a/src/Commands/BackupCommand.php
+++ b/src/Commands/BackupCommand.php
@@ -3,6 +3,7 @@
 namespace Spatie\Backup\Commands;
 
 use Exception;
+use Spatie\Backup\Events\BackupStarted;
 use Spatie\Backup\Events\BackupHasFailed;
 use Spatie\Backup\Exceptions\InvalidCommand;
 use Spatie\Backup\Tasks\Backup\BackupJobFactory;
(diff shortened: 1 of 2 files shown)


introwit:
Hey thanks for the PR :) Could you please also add the tests too?

ctf0:
sorry, i dont know how to write tests, however the pr isnt that big, its just an event that fires b4 the backup start

freekmurze:
Let's make the event more useful by letting it hold a `BackupJob`. Also, fire off the event in the `run` function of `BackupJob`.

ctf0:
@freekmurze am not sure what u r after, why we need the BackupJob ?

freekmurze:
It would be nice of the consumer of the event to know which backup is starting.

ctf0:
actually this was more intended for the client side alert, to inform the user that no writing ops is currently available, but okay, i can make it have the same footprint as BackupWasSuccessful , what u think ?

freekmurze:
Let the event hold `BackupJob`

ctf0:
apologize for the delay, but i found i can achieve what am after through https://laravel.com/docs/5.5/scheduling#task-hooks so no need for this event

PR #606, opened 2018-01-17
Use correct exit codes

mdavis1982:
I'm using this package as part of a deployment hook on Envoyer so that I can backup my site every time I deploy.

Currently, if a backup fails, the deployment script carries on because the `backup:run` command returns an exit code of `-1`.

I've updated it in this PR so that it return `1` instead which signals a general error and will fail the deployment if a backup cannot be taken successfully.


freekmurze:
Thanks!

PR #607, opened 2018-01-18
Rethrow exceptions to exit properly on error

mdavis1982:
This PR contains a couple of things:

## Rethrowing Exceptions
If a Backup or Cleanup task fails, it will re-throw the caught exception, to allow the calling Command to catch the exception again and act accordingly (in this case, returning the correct exit code).

This shouldn't cause any regressions - I've been careful to delete the temporary directory before re-throwing the exception in the `BackupJob` class. The notifications should also still get sent, and the console output should remain the same.

The `run()` methods of both the `BackupJob` and `CleanupJob` are only called in the Commands, and both of these have extra `try..catch` blocks around them to catch the re-thrown exception.

## Fix Failing Tests
I've also fixed the two failing tests that weren't updated in #606. Sorry... That was my mistake 😄


freekmurze:
Thanks!

Issue #614, opened 2018-01-29
Specify which database to backup in --only-db


Corys8646:
Hello, I have a site that has two databases. One database isn't updated that frequently so a backups a few days per week is fine, but the other database is updated more frequently so twice daily is what i'm after.

Is it possibly to specify which database to backup via the command line?

freekmurze:
Currently this is not possible, but I'd accept a PR with tests that adds this to the package.

mastercoder97:
hi good day. i have a question is it possible to backup multiple database in laravel. i have 5 databases to backup. if possible. how? thank you.

freekmurze:
See #679

PR #615, opened 2018-01-30
Allow backup of multiple sqlite database connections

mattdfloyd:
This will mostly be helpful in testing (specifically #614 ).


mattdfloyd:
Rolled into #616

PR #616, opened 2018-01-30
Add ability to specify database connections to be backed up

### LINKED ISSUE #614: Specify which database to backup in --only-db

mattdfloyd:
The `--connection` argument overrides the default configuration. Multiple connections can be specified with multiple `--connection` arguments.

Also adds ability to backup multiple sqlite databases by extracting the name of the sqlite database from the file path. 

Resolves #614 
Closes #615

### DIFF
--- a/src/Commands/BackupCommand.php
+++ b/src/Commands/BackupCommand.php
@@ -10,7 +10,7 @@ use Spatie\Backup\Tasks\Backup\BackupJobFactory;
 class BackupCommand extends BaseCommand
 {
     /** @var string */
-    protected $signature = 'backup:run {--filename=} {--only-db} {--only-files} {--only-to-disk=} {--disable-notifications}';
+    protected $signature = 'backup:run {--filename=} {--only-db} {--connection=*} {--only-files} {--only-to-disk=} {--disable-notifications}';
 
     /** @var string */
     protected $description = 'Run the backup.';
@@ -28,6 +28,10 @@ class BackupCommand extends BaseCommand
 
             if ($this->option('only-db')) {
                 $backupJob->dontBackupFilesystem();
+            }
+
+            if ($this->option('connection')) {
+                $backupJob->setDbDumpers(BackupJobFactory::createDbDumpers($this->option('connection')));
             }
 
             if ($this->option('only-files')) {
--- a/src/Tasks/Backup/BackupJob.php
+++ b/src/Tasks/Backup/BackupJob.php
@@ -214,7 +214,12 @@ class BackupJob
 
             $dbType = mb_strtolower(basename(str_replace('\\', '/', get_class($dbDumper))));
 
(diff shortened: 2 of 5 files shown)


src/Commands/BackupCommand.php
```diff
@@ -24,7 +24,13 @@ public function handle()
         try {
             $this->guardAgainstInvalidOptions();
 
-            $backupJob = BackupJobFactory::createFromArray(config('backup'));
+            $config = config('backup');
+
+            if ($this->option('connection')) {
+                array_set($config, 'backup.source.databases', $this->option('connection'));
```
freekmurze: Instead of overriding the config, I'd rather see some dedicated methods on the  `BackupJobFactory` to set the databases to be backed up.

src/Tasks/Backup/BackupJob.php
```diff
@@ -214,7 +214,13 @@ protected function dumpDatabases(): array
 
             $dbType = mb_strtolower(basename(str_replace('\\', '/', get_class($dbDumper))));
 
-            $dbName = $dbDumper instanceof Sqlite ? 'database' : $dbDumper->getDbName();
+            $dbName = $dbDumper->getDbName();
+
+            if ($dbDumper instanceof Sqlite) {
+                $_ = explode('.', basename($dbName));
```
freekmurze: Could you make it more clear what happens here?

mattdfloyd: Certainly overcomplicated this! Now the assumption is the sqlite database ends with `.sqlite`. We'll use the filename of the database as part of the backup filename.

src/Commands/BackupCommand.php
mattdfloyd: Made `createDbDumpers` on BackupJobFactory public for that logic to be reused.

tests/Integration/BackupCommandTest.php
```diff
@@ -282,4 +282,54 @@ public function it_should_omit_the_backup_failed_event_with_no_notifications_fla
...
+
+        $backupDiskLocal = $this->app['config']->get('filesystems.disks.local.root');
+        $backupFileLocal = $backupDiskLocal . DIRECTORY_SEPARATOR . $this->expectedZipPath;
+        $this->assertFileExistsInZip($backupFileLocal, 'sqlite-database2.sql');
+
+        /*
+         * Close the database connection to unlock the sqlite file for deletion.
+         * This prevents the errors from other tests trying to delete and recreate the folder.
+         */
+        $this->app['db']->disconnect();
```
freekmurze: Put this line and the comments in it's own function.

freekmurze:
Closing due to inactivity.

PR #622, opened 2018-02-06
Added indonesian translation

blissjaspis:
This is translation for Bahasa or Indonesian Language.
thanks before.


freekmurze:
Thanks!

Issue #625, opened 2018-02-08
Set GTID to false when doing mysqldump


necenzurat:
I'm using this to backup a db on Google Cloud SQL.
Trying to restore from backup, it fails. 
The problems is that if a master/slave is set up, the backup is done from the slave. 
Solving this, can be done by adding ```--set-gtid-purged=OFF``` to the mysqldump command.

AlexVanderbist:
Hi, that's a good catch! 

The `mysqldump` command is actually defined in https://github.com/spatie/db-dumper. Feel free to create an issue there or send a PR with a fix 👍.

Once the option for that flag is added in the db-dumper package we can add it in here.

necenzurat:
@AlexVanderbist done, and [merged](https://github.com/spatie/db-dumper/commit/f0f17d12a168d543fa2ba0bdd0cddf68f9dac2f6).

PR #643, opened 2018-03-06
fixes issue #625

necenzurat:


drbyte:
Since this PR is dependent on a newer version release of another package, shouldn't `composer.json` be updated as well?

ie: 
```diff
-        "spatie/db-dumper": "^2.7",
+        "spatie/db-dumper": "^2.9",
```

necenzurat:
ye, @drbyte never saw that @freekmurze did a version bump

drbyte:
The tests are failing due to errors finding/using `setGtidPurged`. Isn't that what this PR was to solve?

freekmurze:
I'm thinking that option isn't so commonly used, so it doesn't deserve a separate key in the config file. Take a look at the docs on how to pass specific options to the db-dumper: https://docs.spatie.be/laravel-backup/v5/installation-and-setup#dumping-the-database

necenzurat:
ah, cool, never saw that, you can ignore this pull request.

Issue #612, opened 2018-01-24
Slack Notifications Incorrect, Google Backup Clean Incorrect


Braunson:
- Laravel 5.5.32
- Package Version: 5.1.5
- Filesystem Package: `nao-pon/flysystem-google-drive`
- Backup to Google

I have Slack notifications turned on for:

- BackupHasFailed
- UnhealthyBackupWasFound
- CleanupHasFailed
- BackupWasSuccessful
- HealthyBackupWasFound
- CleanupWasSuccessful

Commands running are:

- `backup:clean` @ 1am
- `backup:run` @ 2am
- `backup:monitor` @ 3am

Now my issue is the backup is created every day as expected, but the information on Slack about the backup for a few commands is incorrect.

I have a backup every day since the 19th however, each day when the commands run to backup, and this is the error I get when the monitor command runs

```
Spatie Backup Bot APP [3:00 AM]
Important: The backups for MyAcmeApp are unhealthy. There are no backups of this application at all.
Application name: MyAcmeApp
Disk: google
Newest backup size: No backups were made yet
Total storage used: 0 KB
Newest backup date: No backups were made yet
Oldest backup date: No backups were made yet
```

First this isn't true, as there is existing backups already there.

Second error is the successful new backup (the `backup:run` command) does create the backup but reports it as 0 KB on Slack, the other info is also incorrect about there being new previous or latest backups.. this is the Slack notification:

```
Spatie Backup Bot APP [2:00 AM]
Successful new backup!
Application name: MyAcmeApp
Disk: google
Newest backup size: No backups were made yet
Total storage used: 0 KB
Newest backup date: No backups were made yet
Oldest backup date: No backups were made yet
```

Any ideas why the Slack notification information is incorrect and the monitor isn't working properly.

sgallard:
Same with nao-pon/flysystem-google-drive

jpussacq:
In my case the only problem is that it does not **clean in Google** Drive and the disk is getting full. I always receive this message:

> Backup name: 1ibtk0QiroA4yErNjZqncaYQ3oxx5YETl
> Disk: google
> Newest backup size: No backups were made yet
> Total storage used: 0 KB
> Newest backup date: No backups were made yet
> Oldest backup date: No backups were made yet

**Then I have endless backups, because they are never cleaned. The local copies are cleaned well.**

Nota: the backup copies are made correctly, both locally and in Google Drive.

jpussacq:
[...]

PR #646, opened 2018-03-12
filter backup files by mime type when possible

larryx64:
Hey guys!

This pull request is in reference to [issue 612](https://github.com/spatie/laravel-backup/issues/612).

The `backup:monitor` and `backup:clean` commands were not working properly when using `nao-pon/flysystem-google-drive` filesystem.

The issue was due to the fact that the `BackupCollection` class filters files based on their extension, and the Google Drive file system `allFiles` method (used to create the backup collection) returns a list of files named by file ID, and therefore have no extension in their file name.

I thought it could be a good solution to use the `$disk` object that owns the files to check its mime type instead of the extension, when the `$disk` object has the `mimeType` method.

I had to change the `it_can_determine_the_size_of_the_backups` test to set the content of the file to a gzipped string instead, so the mime type would be application/x-gzip. I don't like the fact that I had to change the test to make this work...but it makes sense to me that the content of a zip file should be zipped...

This is my first pull request. I hope I did everything right...please play nice 😊

### DIFF
--- a/src/BackupDestination/BackupCollection.php
+++ b/src/BackupDestination/BackupCollection.php
@@ -9,6 +9,11 @@ class BackupCollection extends Collection
     /** @var null|int */
     protected $sizeCache = null;
 
+    /** @var array */
+    protected static $allowedMimeTypes = [
+        'application/zip', 'application/x-zip', 'application/x-gzip'
+    ];
+
     /**
      * @param \Illuminate\Contracts\Filesystem\Filesystem|null $disk
      * @param array                                            $files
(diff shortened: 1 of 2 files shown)


foremtehan:
The issue still here, I have zip file with older that 7 days 2021-07-03-17-14-36.zip but it doesnt do anything after `backup:clean`

```
Cleaning backups of  on disk local...
Used storage after cleanup: 0 B.
Cleaning backups of  on disk google...
Used storage after cleanup: 3.73 GB.
```

Should i change config file ? i didnt touch anything btw

PR #649, opened 2018-03-15
Fixes in french translation

LEI:
Here are some improvements to the french translation.

I might take it a bit further in another PR and implement translations for the backup destination property keys.


freekmurze:
Merci!

PR #657, opened 2018-04-07
 Setting a temporary directory through the config file

Y2788496:


config/backup.php
```diff
@@ -65,6 +65,9 @@
                 'local',
             ],
         ],
+
+        'temp_directory' => storage_path('app/backup-temp'),
```
freekmurze: Rename this to `temporary_directory`. Also add a sentence explaining the option.

freekmurze:
Closing due to inactivity.

PR #658, opened 2018-04-12
Added compatibility with mariaDB drivers

kikoseijo:
Just to make it work with people using a package like https://github.com/ybr-nx/laravel-mariadb
We are required to change driver name and this leads to errors when trying to make a database backup.

```
'mysql' => [
            'driver' => 'mariadb',
```

It just works same, but with this library we able to query json for your other translatable package.

Thanks.


kikoseijo:
Naaa, thank you!

And thanks for all this great contributions from the Spatie team!

Issue #662, opened 2018-04-24
Remove dumped file after gzip


glamorous:
When you have limited disk space on your local system, you have to have enough space to:

1. Temporary save the mysql dumped file
2. Temporary save the gzipped-version of that dumped file
3. Zip containing (gzipped) dump from database and extra files

When you have a large database, this space requirements will go up quickly. My proposal is to delete the dumped database file after gzipping because you never use it anymore. In only gets deleted after de full cleanup.

See proposal at https://github.com/spatie/laravel-backup/pull/663

PR #663, opened 2018-04-24
#662 - Remove the dumped database file after gzipping it

glamorous:
Like described in #662. This will cleanup disk space when possible before creating zip.


freekmurze:
Thanks!

glamorous:
I'm looking for a better way to dump gzipped mysql. Proposal and PR are coming to db-dumper.

glamorous:
Fix for wrong import: https://github.com/spatie/laravel-backup/pull/666

PR #671, opened 2018-04-27
Database dump is streamed to gzip now, saves a lot of space.

glamorous:
I didn't update the changelog because I wasn't sure to update if the update would be a minor or major improvement. The doubt was because the composer.json file has been changed too, because it uses functionality that wasn't available in a version from db-dumber before the one specified.


freekmurze:
Thanks!

PR #678, opened 2018-05-05
Require php 7.1

dvlpp:
Since the last version, php 7.0 is no longer supported. This PR fix composer.json to prevent php 7.0 apps to DL a non compatible version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         }
     ],
     "require": {
-        "php": "^7.0",
+        "php": "^7.1",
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

sebdesign:
@dvlpp v5.6.5 is not working on PHP 7.0. According to this commit https://github.com/spatie/laravel-backup/commit/b447af7c0c336fd50cfff4cb287cbf0e318c27ac the code has PHP 7.1 syntax, but composer.json does not enforce it, which is causing syntax errors.

PR #699, opened 2018-06-18
Change exclude sources

krzysztofgwiazda:


config/backup.php
```diff
@@ -29,6 +29,8 @@
                 'exclude' => [
                     base_path('vendor'),
                     base_path('node_modules'),
+                    base_path('.git'),
+                    storage_path(),
```
freekmurze: Some people do want to backup their storage dir.

config/backup.php
```diff
@@ -29,6 +29,8 @@
                 'exclude' => [
                     base_path('vendor'),
                     base_path('node_modules'),
+                    base_path('.git'),
```
freekmurze: This is already handled here: https://github.com/spatie/laravel-backup/blob/4d666be08c107c62216f126119d7986cd4456d5f/src/Tasks/Backup/FileSelection.php#L70-L71

PR #707, opened 2018-06-24
Update BackupCommand.php

saidbakr:
Concatenating completion datetime to the complete message "Backup completed!" the parameter of `comment` method for more informative data about time elapsed.

### DIFF
--- a/src/Commands/BackupCommand.php
+++ b/src/Commands/BackupCommand.php
@@ -51,7 +51,7 @@ class BackupCommand extends BaseCommand
 
             $backupJob->run();
 
-            consoleOutput()->comment('Backup completed!');
+            consoleOutput()->comment('Backup completed! @ '.date('Y-m-d H:i:s'));
         } catch (Exception $exception) {
             consoleOutput()->error("Backup failed because: {$exception->getMessage()}.");
 


freekmurze:
Thank you for this. I'm not going to merge it in because most people aren't interested in this change. i'd like to keep that package as simple as possible.  Use one of [these events](https://docs.spatie.be/laravel-backup/v5/taking-backups/events) and the `consoleOutput` function to output anything your heart desires.

saidbakr:
My main idea is when using backup from controller not from artisan command. For example, I, currently, use it from a controller called setting, the user can execute the backup command by accessing an action via PHP `exec` and he/she will be sure that the backup will be carried out even leaving the page, then he/she will look afterward to an output text file with the backup results to be informed about its completion. So, as the output is more informational, it will be tiny and nice tool to inform about performance and to distinguish each backup process.
```php
public function backup(Request $request, $type = 'database')
    {
        try{
           // removing the previous output file.
            @exec("unlink ".base_path().'/public/assets/backuplog/backuplog.txt');        
        }
        catch(\Exception $e){
           "jjj";
        }
        $command0 = 'echo "Started @ '.date('Y-m-d H:i:s').'\n" > '.base_path().'/public/assets/backuplog/backuplog.txt';
        exec($command0);
        $command = 'bash -c "exec nohup setsid php '.base_path().'/artisan backup:run >> '.base_path().'/public/assets/backuplog/backuplog.txt 2>&1 &"';
        try{            
            echo exec($command);           
        }
        catch(\Exception $e){
            dd($e);
        }        
        return '<h1><a href="/assets/backuplog">Checkout the backup log</a></h1>';
    }
```

PR #710, opened 2018-07-04
Add instructions on how to customize the content of database dumps.

mortenscheel:
It took me a while to figure out how to exclude a table from the dump-file, so I added a short instruction on how to customize the dump content.

I'm not sure if this is the right place to add it, or if it's the right level of detail, but I think it needs to be mentioned somewhere. Might be worth mentioning on the documentation page as well.


freekmurze:
Thanks!

PR #718, opened 2018-07-21
Using translation in the BaseNotification class.

wideivan:
I need to translate English to Russian in emails.
So, I added some items to _notifications_ language file and used them in the BaseNotification class.

### DIFF
--- a/resources/lang/en/notifications.php
+++ b/resources/lang/en/notifications.php
@@ -1,6 +1,16 @@
 <?php
 
 return [
+    'application_name' => 'Application name',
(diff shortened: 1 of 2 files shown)


freekmurze:
Thank you for this, but I'm going to leave things as they are for now.

wideivan:
Why?

floreauwe:
Hi ! @freekmurze 
I would like to request the same thing as @ivantorgov.
It can be very handy for us to translate the whole items from the notification language file.

PR #751, opened 2018-09-14
Add bzip2 compress option

RicardoRamirezR:
Adding bzip2 to Laravel-Backup.

> bzip2 creates about 15% smaller files than gzip. bzip2 compresses somewhat slower than gzip, but seems that it hasn't prevented bzip2 from getting popular. 

For quote's source [here](https://tukaani.org/lzma/benchmarks.html)

Edit: bzip2 can be imported with [Sequel Pro](https://sequelpro.com/) as with gzip

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "php": "^7.1",
         "laravel/framework": "~5.5.0|~5.6.0|~5.7.0",
         "league/flysystem": "^1.0.27",
-        "spatie/db-dumper": "^2.10",
+        "spatie/db-dumper": "^2.11.1",
         "spatie/temporary-directory": "^1.1",
         "symfony/finder": "^3.3|^4.0"
     },
--- a/config/backup.php
+++ b/config/backup.php
@@ -62,9 +62,13 @@ return [
         ],
 
         /*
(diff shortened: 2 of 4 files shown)


RicardoRamirezR:
Even more, we can add ```lzma -ez``` if more compression is needed

peter279k:
It looks great :). 
Perhaps we should add the `ext-bz2: "*"` for `require` block in `composer.json`.
This can check whether the `bz2` extension is installed during `composer` does the installation work :+1:.

RicardoRamirezR:
@peter279k Spatie is using the external program ```gzip``` so I added ```bzip2``` as external too.

freekmurze:
Thank you for this. Could you update this PR after you've made the necessary changes to https://github.com/spatie/db-dumper/pull/81

Thanks!

RicardoRamirezR:
Done! but db-dumper should be modified. See [comment#81](https://github.com/spatie/db-dumper/pull/81#issuecomment-424740509)

RicardoRamirezR:
Done.

freekmurze:
Thanks!

PR #772, opened 2018-10-18
Remove drive letter and correct path separator, fixes Windows issues

Turtle1363:
Windows fails when drive letters exist in zip path to file.

This change removes the drive letter if it exist, and sets the path to use only / characters.

This should be backward compatible.  It fixes the test:

tests/Integration/BackupCommandTest it_includes_files_from_the_local_disks_in_the_backup

I've also added a test for determineNameOfFileInZip


freekmurze:
I don't want to maintain regex'es that fix paths. Thank you for this, but leaving it as it is for now.

PR #778, opened 2018-10-29
Update README.md

lucidlemon:
added the oneliner to install the package via composer, moved the "using" part below the "using" headline. removed the duplicate notice of who spatie is since it was kinda weird to have it in the description before the installation part.


freekmurze:
Thank you for this, but we're going to keep things as they are for now.

PR #779, opened 2018-10-29
clear the backup-size cache (fixes #776)

### LINKED ISSUE #776: [Bug] Used storage after cleanup not updated immediately
Hi,

When running the `artisan backup:clean` command, the total backup size shown after "Used storage after cleanup:" is not updated. When running the same command a second time, the total backup size is shown correctly.

Cheers

otherpaco:
when running the `artisan backup:clean` command a wrong backup size is
shown due to caching. Fixes #776


freekmurze:
Very nice, thanks!

Issue #746, opened 2018-09-04
Backup clean fails - file not found


chrisGeonet:
Running backup:clean for the following config will result in an error
Cleanup failed because: File not found at path: backups/site_2017-08-21-16-07-13.zip

Exactly like issue #505 

I am still getting this problem using 5.10.1.
I have two backups 2.6gb in size, and have deleteOldestBackupsWhenUsingMoreMegabytesThan set to 5000.
When i run a cleanup it deletes the oldest backup, then errors saying the same file cannot be found

Jaspur:
Still happening here too 
```
Whoops!

An error occurred while cleaning up the backups of Mijn Bewijsstukken

Exception message: File not found at path: backups-ws/2018-08-31-02-00-03.zip

[...]

PR #782, opened 2018-11-01
Clear cache before attempting to send notification

bencomeau:
This pull request implements a fix for issue #746 and possibly #505.

I believe the issue is related to cached backup destinations -- if all the backup files were to be deleted, expect for the latest, and the `CleanupWasSuccessful` event were to be fired, then an exception would be thrown. This was being caused in the [BaseNotification at line 52](https://github.com/spatie/laravel-backup/blob/master/src/Notifications/BaseNotification.php#L52) when it was trying to grab the `newestBackup` size because it was using cached values.

I have also added a test which will fail if you remove the `$backupDestination->fresh();` introduced in `BaseNotification` within this PR.

### DIFF
--- a/src/Notifications/BaseNotification.php
+++ b/src/Notifications/BaseNotification.php
@@ -39,6 +39,8 @@ abstract class BaseNotification extends Notification
             return collect();
         }
 
+        $backupDestination->fresh();
+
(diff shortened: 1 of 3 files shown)


freekmurze:
Thanks!

PR #785, opened 2018-11-08
Slack Webhook URL config in config null issue

xiomai:
Avoid setting the variable in `config()` for it will return null value.
Use `env()` helper instead, or string of the url.

`src\Notifications\Notifiable.php@routeNotificationForSlack` is
already using `config()` helper.
Reference of the issue: 
https://blog.maqe.com/dont-use-laravel-s-config-inside-config-files-40e2c8207225

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -124,6 +124,10 @@ return [
         ],
 
         'slack' => [
+            /*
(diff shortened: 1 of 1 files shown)


freekmurze:
Passing on this for now. It's not the config file responsability to teach users not to use `config` but `env`.

PR #789, opened 2018-11-09
Custom health checks

rasmuscnielsen:
This PR introduces the concept of health inspections.

**Concept**

- A health inspection can be added to the `backup.php` config, and will be run on each `backup:monitor`.
- A health inspection can be any class added that extends the abstract `HealthInspection` class.
- Configuration looks like the following
```php
'monitorBackups' => [
    [
        'name' => config('app.name'),
        'disks' => ['local'],
        'newestBackupsShouldNotBeOlderThanDays' => 1,
        'storageUsedMayNotBeHigherThanMegabytes' => 5000,
        'inspections' => [
            SomeInspection::class, // default configuration
            SomeInspection::class => ['option' => 'value'], // with custom configuration
        ],
    ],
]
```

**Considerations for the future**

With the introduction of inspections the existing health checks in `BackupDestinationStatus` could be refactored into inspections and help decouple `BackupDestinationStatus` and `Notifications\UnhealthyBackupWasFound`. 

**Example use cases**

A couple ideas for helpful inspections could be 

- Always increasing file size: if the newest backup has greatly reduced size than the previous, raise a warning. This would help catch situations where a lot of files were deleted or moved and backup config not amended. Or DB dumps started failing. **This is already implemented in a new feature branch and ready to ship if this PR gets accepted.** [See implementation here](https://github.com/makeabledk/laravel-backup/compare/feature/custom-health-inspections...makeabledk:feature/increasing-filesize-inspection?expand=1)  
- Validate DB dumps: spin up the latest DB dump in a new database and make some health checks

The great thing is that the package would be easily extendibly with custom inspections suited for the individual project.

____

Feedback and suggestions are appreciated


freekmurze:
Thank you for submitting this. I really like the proposed functionality. I got two big remarks:

1. I don't like the `inspection` naming. I think everything should revolve our a `HealthCheck` and all classes, functions, ... should be named like that. 

2. I'd like a `HealthCheck` to become a first class citizen of the package. The current `newestBackupsShouldNotBeOlderThanDays` and `storageUsedMayNotBeHigherThanMegabytes` should be converted to each their own `HealthCheck` class. It should be possible to pass options to those `HealthCheck` classes from the config file.

```
    'monitor_backups' => [
        [
            'name' => config('app.name'),
            'disks' => ['local'],
            'healthChecks' = > [
                 Spatie\Backup\Tasks\Monitor\HealthChecks\MaximumAgeInDays::class => 1,
                 Spatie\Backup\Tasks\Monitor\HealthChecks\MaximumStorageInMegabytes::class => 5000,
           ],
        ],
```

Of course this will become a breaking change, but I'm willing to tag a new major version for this feature.

freekmurze:
I've changed the base of this PR to v6. In that version all config keys are snake_cased. Could you rebase your branch to v6?

rasmuscnielsen:
@freekmurze Thanks for the feedback and glad to hear you like the proposal.

I've rebased to v6 and did some work on the implementation:

- Renamed inspections to HealthCheck
- Extracted 'reachable', 'max age' and 'max storage' checks to each their own HealthCheck. This also cleaned up the `UnhealthyBackupWasFound` notification 
- Added tests to health checks. 
- Implemented the proposed config syntax: `MaximumAgeInDays::class => 1`.
- Refactored the `ListCommand`

Please have another look and let me know in case of questions

riyuk:
I would love to see this `Validate DB dumps: spin up the latest DB dump in a new database and make some health checks` in v6.
It's still strange that there is no command to simply write all dumps to a db.

Could be useful if you have several test-systems and want to install the latest dumps on that system.
Also it would be great if you can target specific tables u want to import from that backup files. ("do not import users from live-system; just use THAT data")

rasmuscnielsen:
@riyuk In fact this is already under active development in another repo of ours: https://github.com/makeabledk/laravel-backup-sql-check

However as this is quite comprehensive it's more a 'scratch our own itch' type of thing. So for our own use case we'll only be implementing MySQL driver in the foreseeable future.

Basically it's a matter of re-implementing https://github.com/spatie/db-dumper as the reversed 'db-importer'. Then you can build your restoration feature on top of that.
This is what we've attempted to do in this WIP repo for MySQL only so far, and only for the purpose of spinning up the DB dump only to delete it straight away.

src/Tasks/Monitor/HealthChecks/MaximumStorageInMegabytes.php
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Spatie\Backup\Tasks\Monitor\HealthChecks;
+
+use Spatie\Backup\Helpers\Format;
+use Spatie\Backup\Tasks\Monitor\HealthCheck;
+use Spatie\Backup\BackupDestination\BackupDestination;
+
+class MaximumStorageInMegabytes extends HealthCheck
+{
+    protected $allowance;
```
freekmurze: Let's rename this to something more meaningful like `maximumSizeInMegaBytes`. I don't care that it's a long variable name.

src/Tasks/Monitor/HealthChecks/MaximumStorageInMegabytes.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $this->failIf($this->exceedsAllowance($usage = $backupDestination->usedStorage()),
+            trans('backup::notifications.unhealthy_backup_found_full', [
+                'disk_usage' => $this->humanReadableSize($usage),
+                'disk_limit' => $this->humanReadableSize($this->bytes($this->allowance)),
+            ])
+        );
+    }
+
+    protected function exceedsAllowance($usage)
```
freekmurze: Add a return type hint here.

src/Tasks/Monitor/HealthChecks/MaximumStorageInMegabytes.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        return $usage > $this->bytes($this->allowance);
+    }
+
+    protected function bytes($mb): int
+    {
+        return (int) $mb * 1024 * 1024;
+    }
+
+    protected function humanReadableSize($sizeInBytes): string
```
freekmurze: Add a type hint here

config/backup.php
```diff
@@ -147,16 +147,14 @@
         [
             'name' => config('app.name'),
             'disks' => ['local'],
-            'newest_backups_should_not_be_older_than_days' => 1,
-            'storage_used_may_not_be_higher_than_megabytes' => 5000,
+            'health_checks' => [],
```
freekmurze: Add those three health checks here

src/Commands/ListCommand.php
```diff
@@ -20,8 +21,7 @@ public function handle()
         $statuses = BackupDestinationStatusFactory::createForMonitorConfig(config('backup.monitor_backups'));
 
         $this->displayOverview($statuses);
```
freekmurze: Return `$this` in those methods so you can chain them here.

src/Tasks/Monitor/BackupDestinationStatus.php
```diff
@@ -2,158 +2,64 @@
...
 
-    public function newestBackupIsTooOld(): bool
+    public function check(HealthCheck $check)
     {
-        if (! count($this->backupDestination->backups())) {
-            return true;
+        try {
+            $check->handle($this->backupDestination());
+        } catch (\Exception $exception) {
+            return new HealthCheckFailure($check, $exception);
```
freekmurze: Create a static methods on that exception, so we don't need to new it up.

More info: http://rosstuck.com/formatting-exception-messages

src/Tasks/Monitor/HealthChecks/MaximumStorageInMegabytes.php
```diff
@@ -0,0 +1,46 @@
...
+            ])
+        );
+    }
+
+    protected function exceedsAllowance($usage)
+    {
+        return $usage > $this->bytes($this->allowance);
+    }
+
+    protected function bytes($mb): int
```
freekmurze: add a type hint here, rename the parameter to `$megaBytes`

config/backup.php
rasmuscnielsen: In fact `IsReachable` is prepended automatically in `BackupDestinationStatus` as it is a quite essential. No other checks makes sense if it is not reachable.

So I've just added the 2 other checks with their default config

src/Tasks/Monitor/BackupDestinationStatus.php
rasmuscnielsen: In fact this is not an exception itself but a wrapper around the exception with a bunch of help methods to interact with the health check and the failure.

I'm not sure how to apply the tactics in this case?

rasmuscnielsen:
@freekmurze pushed another version and replied to a few of your comments

config/backup.php
freekmurze: Perfect!

freekmurze:
Pulling this in so I can polish the feature and v6 a bit myself.

Thank you very much for this!

PR #793, opened 2018-11-22
fix lumen compatibility for dependency schema

BePsvPT:
This pull request revert the dependency schema to make it lumen compatible.


freekmurze:
Thanks!

PR #801, opened 2018-12-05
[v6] Add new health check: IncreasingFileSize

rasmuscnielsen:
This PR introduces a `IncreasingFileSize` health check.

If a backup reduces in size beyond a certain tolerance from the previous backup, it will raise a health warning.

This is useful if a large amount of files was accidentally deleted or moved outside the scope of the backup script.


freekmurze:
Hi, that's a nice check

Because I'm not going to use this myself, I don't want to maintain it. 

Could you put your code in a separate repo. i'll happily link to it from our docs.

PR #809, opened 2018-12-14
add extra options for writeStream

yushkevichv:
Add optionally params for S3-compability storages. For example, you can set storageClass.

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -87,6 +87,13 @@ return [
             'disks' => [
                 'local',
             ],
+
+            /*
+             *  The extra options for set file visibility, storageClass, etc
+             */
+            'extra_options' => [
+                //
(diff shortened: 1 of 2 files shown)


freekmurze:
Thank you for this.

Wouldn't those options be disk specific?

Instead of adding those options in our config file, I'd much rather have the extra option added in the `filesystems` config file. Much like we do for dumping the database: we let the user add a `dump` key.
https://docs.spatie.be/laravel-backup/v5/installation-and-setup#dumping-the-database

Closing this for now, I'd welcome a new PR that adds this as described above, and with tests.

yushkevichv:
Yes, I agree with you. This feature is disk specific. I change config source and send new PR. 
Thank you for your feedback.

PR #811, opened 2018-12-18
add extra options to backup destination at writeStream

yushkevichv:
Add optionally params for S3-compability storages. 
For usage you should set at config/filesystems.php for your backup disk array options 'dump_extra_options', for example:
```
's3-backup' => [
            'driver' => 's3',
             ...,

            'backup_extra_options' => [
                'StorageClass' => 'COLD'
            ]
        ],
```

### DIFF
--- a/src/BackupDestination/BackupDestination.php
+++ b/src/BackupDestination/BackupDestination.php
@@ -117,7 +117,7 @@ class BackupDestination
 
     public function extraOptions(): ?array
     {
-        $extraConfig = config('filesystems.disks.'.$this->diskName().'.dump_extra_options') ?? [];
+        $extraConfig = config('filesystems.disks.'.$this->diskName().'.backup_extra_options') ?? [];
 
         if (! is_array($extraConfig) || (count($extraConfig) < 1)) {
             return [];
--- a/src/BackupDestination/BackupDestination.php
+++ b/src/BackupDestination/BackupDestination.php
@@ -85,7 +85,7 @@ class BackupDestination
         $options = $this->extraOptions();
 
         $this->disk->getDriver()->writeStream($destination, $handle, $options);
-        
+
         if (is_resource($handle)) {
             fclose($handle);
         }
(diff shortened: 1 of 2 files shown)


src/BackupDestination/BackupDestination.php
```diff
@@ -113,6 +115,17 @@ public function connectionError(): Exception
         return $this->connectionError;
     }
 
+    public function extraOptions(): ?array
+    {
+        $extraConfig = config('filesystems.disks.'.$this->diskName().'.dump_extra_options') ?? [];
```
freekmurze: As this does not concern the database dump this option should be renamed. What about `backup_extra_options`?

yushkevichv: About rename - no problem.  

As for the tests, I'll think about which ones are better to add. Maybe you will have ideas?
Since the config is taken from filesystems,  I think should be need to add a factory and refactor some code more globally.

I will write some tests in the coming days.

src/BackupDestination/BackupDestination.php
```diff
@@ -113,6 +115,17 @@ public function connectionError(): Exception
         return $this->connectionError;
     }
 
+    public function extraOptions(): ?array
+    {
+        $extraConfig = config('filesystems.disks.'.$this->diskName().'.backup_extra_options') ?? [];
+
+        if (! is_array($extraConfig) || (count($extraConfig) < 1)) {
```
freekmurze: It seems like this check isn't necessary anymore, because you already default it to an empty array.

src/BackupDestination/BackupDestination.php
```diff
@@ -113,6 +115,17 @@ public function connectionError(): Exception
         return $this->connectionError;
     }
 
+    public function extraOptions(): ?array
```
freekmurze: This will always return an array.

freekmurze:
Seems almost ready to be pulled in 👍 Added a few comments...

freekmurze:
I'll fix those last issues myself. Thank you for your work on this 👍

PR #823, opened 2018-12-28
Added Dutch translation

DenzoNL:
Hey all, I noticed a Dutch translation was missing since my app uses the `nl` locale, so I created one. Some notes on the translation as I am not a professional translator;

- I've used the term 'niet gezond' rather than 'ongezond' for unhealthy as I felt the latter could be more easily read as 'gezond' while skimming through the notifications. I'm sure there is a better way of phrasing it though.

- The grammatically correct translation for 'backup destination' is 'back-upbestemming' and I'm currently using that, but it's not very comfortable to read.

- I've mostly translated the English sentences 1:1 including the sentence structures, so there could be some anglicism in there.

It's more than a decent enough translation for its use case, but feel free to improve on it.

### DIFF
--- a/resources/lang/nl/notifications.php
+++ b/resources/lang/nl/notifications.php
@@ -1,0 +1,35 @@
+<?php
+
+return [
+    'exception_message' => 'Exceptie bericht: :message',
+    'exception_trace' => 'Exceptie trace: :trace',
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you for this, Dutch is my mother tongue, so I'll do some small fixes.

PR #837, opened 2019-01-01
Updates monitor_backups.name config value

mvdnbrk:
In v6.0 the config value of `backup.name` changed from `config('app.name')` to `env('APP_NAME')`.

This PR makes the same change to  the `monitor_backups.name` config value.


freekmurze:
Thanks!

PR #845, opened 2019-01-10
Fixed mongo user auth error + fix wrong extension for mongoDB export

Giacomo92:


config/backup.php
```diff
@@ -60,7 +60,7 @@
                 'mysql',
             ],
         ],
-
+        'mongodb_user_auth' => env('MONGO_DB_AUTH', 'admin'),
```
freekmurze: I'm thinking we should add a new config value, but use the `dump` config key on a database connection. Example here: https://docs.spatie.be/laravel-backup/v6/installation-and-setup#dumping-the-database

freekmurze: I don't see a reason why `mongodb_user_auth` should be in `config/backup.php` and can't live as extra option in the dump config like mentioned here: https://docs.spatie.be/laravel-backup/v6/installation-and-setup#dumping-the-database

src/Tasks/Backup/BackupJob.php
```diff
@@ -229,7 +230,12 @@ protected function dumpDatabases(): array
 
             $dbName = $dbDumper instanceof Sqlite ? 'database' : $dbDumper->getDbName();
 
-            $fileName = "{$dbType}-{$dbName}.sql";
+	        if ($dbDumper instanceof MongoDb) {
+		        $fileName = "{$dbType}-{$dbName}.archive";
+	        }else{
```
freekmurze: I 'don't like `else` statements. Please refactor this to a ternary.

config/backup.php
Giacomo92: I fix what you ask @freekmurze. I have also added some test with mongo, but how can i fake a mongodb connection like you do in function [`createSQLiteDatabase`](https://github.com/spatie/laravel-backup/blob/25571aeb16ba8579b8426c96a9c4134394502be3/tests/TestCase.php#L159)?

src/Tasks/Backup/BackupJob.php
```diff
@@ -2,276 +2,256 @@
...
+	protected function dumpDatabases(): array {
+		return $this->dbDumpers->map( function ( DbDumper $dbDumper ) {
+			consoleOutput()->info( "Dumping database {$dbDumper->getDbName()}..." );
 
-            $dbName = $dbDumper instanceof Sqlite ? 'database' : $dbDumper->getDbName();
+			$dbType = mb_strtolower( basename( str_replace( '\\', '/', get_class( $dbDumper ) ) ) );
 
-            $fileName = "{$dbType}-{$dbName}.sql";
+			$dbName   = $dbDumper instanceof Sqlite ? 'database' : $dbDumper->getDbName();
+			$fileName = $dbDumper instanceof MongoDb ? "{$dbType}-{$dbName}.archive" : "{$dbType}-{$dbName}.sql";
```
freekmurze: Instead of having a special case for `MongoDb` introduce a new function `getExtension` that returns the extension to use.

src/Tasks/Backup/BackupJob.php
```diff
@@ -2,276 +2,256 @@
...
-    {
-        $this->dontBackupFilesystem();
-        $this->dontBackupDatabases();
-        $this->setDefaultFilename();
+	/** @var bool */
+	protected $sendNotifications = true;
 
-        $this->backupDestinations = new Collection();
-    }
+	public function __construct() {
```
freekmurze: Please revert all these formatting changes, the make the diff hard to read.

Giacomo92:
@freekmurze any update on this?

freekmurze:
Thanks!

PR #858, opened 2019-02-01
Use Str:: & Arr:: Instead Of Helper Methods

Te7a-Houdini:


ce-brex:
The `starts_with` helper is also being deprecated and should be switched out with `Str::startsWith()`.

- src/Tasks/Backup/Zip.php:40
- src/Tasks/Backup/FileSelection.php:114

Te7a-Houdini:
@ce-brex thanks for the catch

freekmurze:
Thank you! 👍

PR #872, opened 2019-02-27
Change amount to number

danielsdeboer:
I realise this is a bonkers tiny nitpick... I get to see this notification multiple times per day and it always bugs me just a little bit.


freekmurze:
Thanks!

PR #877, opened 2019-03-01
Add support for Laravel 5.8

pawelmysior:
For those of us who still work on PHP 7.1.* but want to use Laravel 5.8 :)


freekmurze:
Thanks

PR #889, opened 2019-04-02
fix bug multiple sqlite dump-file is same name

zokei-oba:

### DIFF
--- a/src/Tasks/Backup/BackupJob.php
+++ b/src/Tasks/Backup/BackupJob.php
@@ -222,12 +222,15 @@ class BackupJob
      */
     protected function dumpDatabases(): array
     {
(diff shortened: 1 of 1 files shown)


freekmurze:
The tests seems to be failing? Could you rebase this against the current master? Than might fix it.

zokei-oba:
Thank you for review it.

The test is correct failing.
I understood that the sqlite dump file is named 'sqlite-database.sql' on test too.
Sorry for my pull request with failed test. (The cause I missed was that it failed at the other point, when run at windows...)
I 'll solve this problem and try to post next PR.

I would like to apply this patch to v5. Do I need to pull requests to target the master first?

PR #890, opened 2019-04-02
improve method statement.

nikafzar:


freekmurze:
Thanks, but I prefer the current code.

PR #892, opened 2019-04-04
fix bug multiple sqlite dump-file is same name. (Retry)

zokei-oba:

(retry of #889)


zokei-oba:
I do not know why the test (except for PHP 7.1) fails.
Next time do I need to rebase this to the current master?

freekmurze:
I see that this PR targets v5. We don't maintain that version anymore. Could you target this at v6?

zokei-oba:
Thank you for merge and release for v6. I will use this, someday when I upgrade Laravel in my app.
I need to use v5 for a while, so I use the forked version.

PR #895, opened 2019-04-04
fix bug multiple sqlite dump-file is same name. (v6)

zokei-oba:

(This PR is same as #892. target to v6.)


freekmurze:
Thanks!

PR #900, opened 2019-04-25
Add support to add custom dumpers to DbDumperFactory

patrickbrouwers:
This feature adds a way to add custom database dumpers. In my case this is useful, because we are dealing with a docker container and we cannot rely on the mysql dump command outside that container (different mysql versions unfortunately). 

You can either add dumpers for custom connections:

```
DbDumperFactory::extend('some_other_connection', function() {
    return new YourCustomDumper();
});
```

Or override an existing dumper:

```
DbDumperFactory::extend('mysql', function() {
    return new YourCustomMysqlDumper();
});
```

Relevant docs update: https://github.com/spatie/docs.spatie.be/pull/411

### DIFF
--- a/src/Tasks/Backup/DbDumperFactory.php
+++ b/src/Tasks/Backup/DbDumperFactory.php
@@ -13,6 +13,8 @@ use Spatie\Backup\Exceptions\CannotCreateDbDumper;
 
 class DbDumperFactory
 {
+    protected static $custom = [];
+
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

wadimsewo:
@patrickbrouwers can you please tell me, where I should place this extend code, to an add custom dumper?

patrickbrouwers:
@wadimsewo I've placed it in the boot method of a service provider

Issue #899, opened 2019-04-24
Password protect zip archive


Ruud-Gerrits:
Is it possible to password protect the zip? Maybe with using the setEncryptionName method in ZipArchive (https://www.php.net/manual/en/ziparchive.setencryptionname.php)?

freekmurze:
The package doesn't do this by default.

You can listen for the `BackupZipWasCreated` event  to do last minute manipulations on the created zip file.

https://docs.spatie.be/laravel-backup/v6/taking-backups/events#backupzipwascreated

PR #901, opened 2019-04-25
Added password protecting zip feature

Ruud-Gerrits:
This feature adds a option to password protect a zip with the Ziparchive::setEncryptionName. The password can be set in the config. If the password is null or empty the zip will not be protected.

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -10,6 +10,11 @@ return [
          */
         'name' => env('APP_NAME', 'laravel-backup'),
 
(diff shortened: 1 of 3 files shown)


freekmurze:
Thank you for this, but we're not going to merge it. We want to keep this package as simple as possible, and we don't need this feature.

Please use this solution: https://github.com/spatie/laravel-backup/issues/899#issuecomment-486367175

Issue #904, opened 2019-05-10
PHP version


mvdnbrk:
The current version  of `spatie/laravel-backup` requires PHP version 7.2.

However, spatie/db-dumper 2.14.1 requires PHP 7.3.
This results in a conflict:

```
- Installation request for spatie/db-dumper 2.14.1 -> satisfiable by spatie/db-dumper[2.14.1].
- spatie/db-dumper 2.14.1 requires php ^7.3 -> your PHP version (7.2.18) does not satisfy that requirement.
```

freekmurze:
Either upgrade to PHP 7.3 or require the last PHP 7.2 compatible version of db-dumper. 

Tbh I expected composer to just pick the right versions for you

mvdnbrk:
Upgrading to PHP 7.3 is not a solution for everyone ;)

mvdnbrk:
Anyone who has pulled in this package with composer:

```javascript
"spatie/laravel-backup": "^6.0"
```

now should require the `spatie/db-dumper` v2.14 manually, like so:

```bash
composer require spatie/db-dumper:2.14
```

This might be helpful to anyone else.
Thanks anyway!

PR #905, opened 2019-05-10
Updates composer.json.

mvdnbrk:
See #904

Would suggest releasing this as a patch, and releasing a minor/major version for PHP 7.3 compatability.


freekmurze:
Going to leave it like it is, because the package can work with the older version of db-dumper.

PR #917, opened 2019-05-29
Allows different backup directory names between disks

oriceon:
In some situations, the name of the Backup folder is different from the name generally defined in config, name section.

Therefore, I thought the name could be string (default as before) or array of 'disk name' => 'Backup folder name' to be able to insert the name of the backup folder, let`s say from google drive eq: 2NDOVItvm9aCGTYCBiaRHYlr_P2k1QBbJ.

So, to work with google drive also, you should have config like:

```php
'default_name' => 'Backups',

'name' => [
    'google' => env('FILESYSTEM_DISKS_GOOGLE_DRIVE_BACKUPS_FOLDER_ID'),
],
```

### DIFF
--- a/resources/lang/ro/notifications.php
+++ b/resources/lang/ro/notifications.php
@@ -1,0 +1,35 @@
+<?php
+
+return [
+    'exception_message' => 'Cu excepția mesajului: :message',
+    'exception_trace' => 'Urmă excepţie: :trace',
+    'exception_message_title' => 'Mesaj de excepție',
+    'exception_trace_title' => 'Urmă excepţie',
+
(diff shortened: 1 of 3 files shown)


freekmurze:
I've decided not to add this in for now. Could you create a new PR with only those translations?

Thanks!

oriceon:
What's wrong with what I'm proposing?

freekmurze:
There's nothing really wrong with it (although tests and documentation were missing), but I like to keep this package small. I only add things that I need myself or want to maintain.

Feel free to fork the package and maintain your own customized version.

PR #921, opened 2019-06-11
Add example on how to use useSingleTransaction to avoid locking DB's

SDekkers:
I was looking on how to implement this option myself, as I backup DB's that are ~3GB in size, and with locking the tables it takes a while before they're available again.

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -53,6 +53,7 @@ return [
              *            ]
              *       ]
              * ],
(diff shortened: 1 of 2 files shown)


freekmurze:
Could you transform this to a separate note?

SDekkers:
Added it as separate note

freekmurze:
Thanks

Issue #783, opened 2018-11-04
HEAD - 503 Slow Down


repat:
> Error executing "HeadObject" on "https://xxx.sfo2.digitaloceanspaces.com/xxx/xxx-10-17-02-00-02.zip"; AWS HTTP error: Server error: `HEAD https://xxx.sfo2.digitaloceanspaces.com/xxx/xxx2018-10-17-02-00-02.zip` resulted in a `503 Slow Down` response  (server): 503 Slow Down -

related: #618

Any fixes so far? @mdavis1982 @freekmurze @leeuwd

freekmurze:
I'm thinking this problem should be solved in the FlySystem driver. Either by providing some kind of throttling functionality in the S3 driver or a dedicated Spaces driver that has the right throttling baked in.

pehbehbeh:
@repat Did you come up with a solution? We are also experiencing this issue with Digital Ocean Spaces when executing the cleanup command. Would be happy to help.

> Error executing "DeleteObject" on "https://xxx.digitaloceanspaces.com/xxx/xxx/xxx.zip&quot;; AWS HTTP error: Server error: DELETE https://xxx.digitaloceanspaces.com/xxx/xxx/xxx.zip resulted in a 503 Slow Down response: <?xml version="1.0" encoding="UTF-8"?> <Error> <Code>SlowDown</Code> <Message>Please reduce your request rate.</Message> (truncated...) SlowDown (server): Please reduce your request rate. - <?xml version="1.0" encoding="UTF-8"?> <Error> <Code>SlowDown</Code> <Message>Please reduce your request rate.</Message> <RequestId></RequestId> </Error>

```
#0 /app/vendor/aws/aws-sdk-php/src/WrappedHttpHandler.php(97): Aws\WrappedHttpHandler->parseError(Array, Object(GuzzleHttp\Psr7\Request), Object(Aws\Command), Array)
#1 /app/vendor/guzzlehttp/promises/src/Promise.php(203): Aws\WrappedHttpHandler->Aws{closure}(Array)
#2 /app/vendor/guzzlehttp/promises/src/Promise.php(174): GuzzleHttp\Promise\Promise::callHandler(2, Array, Array)
#3 /app/vendor/guzzlehttp/promises/src/RejectedPromise.php(40): GuzzleHttp\Promise\Promise::GuzzleHttp\Promise{closure}(Array)
[...]

Issue #618, opened 2018-02-01
Doesn't Return Correct Exit Code in Some Situations


mdavis1982:
We're using this package as part of a Deployment Hook in Envoyer to take a backup of our site before a deploy happens.

We did some work in #606 and #607 to help return the correct exit codes in most situations so that our deployment would fail if the backup did not complete successfully.

We're backing up to a Digital Ocean Spaces account (which is S3 compatible). 

Sometimes, however, when a backup is run, we get an exception with a `503 Slow Down` response. The package notifies us via Slack that the backup failed, but the deployment script still runs through to completion.

So, in this situation, the CLI commands are still returning an incorrect exit code.

mdavis1982:
In case it's useful, here is the Exception we get and the Exception trace:

### Exception message
```
Error executing "HeadObject" on "https://redacted.ams3.digitaloceanspaces.com/Backups/Redacted/redacted-backup-2018-01-24-02-00-01.zip"; AWS HTTP error: Server error: `HEAD https://redacted.ams3.digitaloceanspaces.com/Backups/Redacted/redacted-backup-2018-01-24-02-00-01.zip` resulted in a `503 Slow Down` response  (server): 503 Slow Down -
```

### Exception trace
```
#0 /home/forge/redacted.com/releases/20180131105054/vendor/aws/aws-sdk-php/src/WrappedHttpHandler.php(100): Aws\WrappedHttpHandler->parseError(Array, Object(GuzzleHttp\Psr7\Request), Object(Aws\Command), Array)
#1 /home/forge/redacted.com/releases/20180131105054/vendor/guzzlehttp/promises/src/Promise.php(203): Aws\WrappedHttpHandler->Aws\{closure}(Array)
#2 /home/forge/redacted.com/releases/20180131105054/vendor/guzzlehttp/promises/src/Promise.php(174): GuzzleHttp\Promise\Promise::callHandler(2, Array, Array)
#3 /home/forge/redacted.com/releases/20180131105054/vendor/guzzlehttp/promises/src/RejectedPromise.php(40): GuzzleHttp\Promise\Promise::GuzzleHttp\Promise\{closure}(Array)
#4 /home/forge/redacted.com/releases/20180131105054/vendor/guzzlehttp/promises/src/TaskQueue.php(47): GuzzleHttp\Promise\RejectedPromise::GuzzleHttp\Promise\{closure}()
[...]

PR #927, opened 2019-06-25
Prevent excessive IO calls by caching Backup properties

orottier:
Hi, I whipped this up quickly after running into S3 rate limit issues (see also #783 #618)

The number of calls made to the storage backend seemed a bit high to me, and by caching these properties I managed to reduce them 4-fold.

I hope this is an acceptable patch to you. Of course caching these properties makes you prone to some invalid states such as running deletes outside the knowledge of the `Backup` file, but as far as I can see this is currently not happening.

### DIFF
--- a/src/BackupDestination/Backup.php
+++ b/src/BackupDestination/Backup.php
@@ -47,7 +47,6 @@ class Backup
     {
         if ($this->date === null) {
             $this->date = Carbon::createFromTimestamp($this->disk->lastModified($this->path));
-            $this->exists = true;
         }
 
         return $this->date;
@@ -64,7 +63,6 @@ class Backup
             }
 
             $this->size = $this->disk->size($this->path);
-            $this->exists = true;
         }
 
         return $this->size;
@@ -78,8 +76,6 @@ class Backup
     public function delete()
     {
         $this->exists = null;
-        $this->date = null;
-        $this->size = null;
 
         $this->disk->delete($this->path);
 
--- a/src/BackupDestination/Backup.php
+++ b/src/BackupDestination/Backup.php
@@ -13,6 +13,15 @@ class Backup
     /** @var string */
     protected $path;
 
+    /** @var bool */
+    protected $exists;
+
+    /** @var Carbon */
+    protected $date;
+
+    /** @var int */
+    protected $size;
+
     public function __construct(Filesystem $disk, string $path)
     {
         $this->disk = $disk;
@@ -27,12 +36,21 @@ class Backup
 
     public function exists(): bool
     {
-        return $this->disk->exists($this->path);
+        if ($this->exists === null) {
+            $this->exists = $this->disk->exists($this->path);
+        }
+
+        return $this->exists;
     }
 
     public function date(): Carbon
     {
-        return Carbon::createFromTimestamp($this->disk->lastModified($this->path));
+        if ($this->date === null) {
+            $this->date = Carbon::createFromTimestamp($this->disk->lastModified($this->path));
+            $this->exists = true;
+        }
+
+        return $this->date;
     }
 
     /**
@@ -40,11 +58,16 @@ class Backup
      */
     public function size(): int
     {
-        if (! $this->exists()) {
-            return 0;
+        if ($this->size === null) {
+            if (! $this->exists()) {
+                return 0;
+            }
+
+            $this->size = $this->disk->size($this->path);
+            $this->exists = true;
         }
 
-        return $this->disk->size($this->path);
+        return $this->size;
     }
 
     public function stream()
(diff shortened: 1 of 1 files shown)


src/BackupDestination/Backup.php
```diff
@@ -27,24 +36,38 @@ public function path(): string
...
+
+        return $this->exists;
     }
 
     public function date(): Carbon
     {
-        return Carbon::createFromTimestamp($this->disk->lastModified($this->path));
+        if ($this->date === null) {
+            $this->date = Carbon::createFromTimestamp($this->disk->lastModified($this->path));
+            $this->exists = true;
```
freekmurze: Do we really need to set `exists` here?

src/BackupDestination/Backup.php
```diff
@@ -27,24 +36,38 @@ public function path(): string
...
 
     /**
      * Get the size in bytes.
      */
     public function size(): int
     {
-        if (! $this->exists()) {
-            return 0;
+        if ($this->size === null) {
+            if (! $this->exists()) {
```
freekmurze: Do we really need to set `exists` here? This code would be much simpler if we only think of the  `size` here.

src/BackupDestination/Backup.php
```diff
@@ -54,6 +77,10 @@ public function stream()
 
     public function delete()
     {
+        $this->exists = null;
```
freekmurze: Do we really need to set these to `null`. I assume that when calling `delete`, this instance won't be used anymore by the rest of the code.

src/BackupDestination/Backup.php
orottier: I think so yes. This will prevent extra calls. E.g. the IO operations when running `it_can_remove_old_backups_until_using_less_than_maximum_storage` are:
```date   mysite/test1.zip
date   mysite/test2.zip
date   mysite/test3.zip
date   mysite/test4.zip
date   mysite/test2.zip
date   mysite/test3.zip
date   mysite/test4.zip
date   mysite/test2.zip
date   mysite/test3.zip
date   mysite/test4.zip
date   mysite/test2.zip
date   mysite/test3.zip
date   mysite/test4.zip
date   mysite/test2.zip
date   mysite/test3.zip
date   mysite/test4.zip
exists mysite/test2.zip
date   mysite/test2.zip
exists mysite/test3.zip
date   mysite/test3.zip
exists mysite/test4.zip
date   mysite/test4.zip
exists mysite/test2.zip
exists mysite/test3.zip
exists mysite/test4.zip
size   mysite/test2.zip
exists mysite/test2.zip
size   mysite/test3.zip
exists mysite/test3.zip
size   mysite/test4.zip
exists mysite/test4.zip
size   mysite/test1.zip
exists mysite/test1.zip
exists mysite/test2.zip
exists mysite/test3.zip
exists mysite/test4.zip
exists mysite/test2.zip
exists mysite/test3.zip
size   mysite/test2.zip
exists mysite/test2.zip
size   mysite/test3.zip
exists mysite/test3.zip
size   mysite/test1.zip
exists mysite/test1.zip
exists mysite/test2.zip
exists mysite/test3.zip
exists mysite/test2.zip
size   mysite/test2.zip
exists mysite/test2.zip
size   mysite/test1.zip
exists mysite/test1.zip
date   mysite/test1.zip
date   mysite/test2.zip
exists mysite/test1.zip
exists mysite/test2.zip
size   mysite/test1.zip
exists mysite/test1.zip
size   mysite/test1.zip
exists mysite/test1.zip
size   mysite/test2.zip
exists mysite/test2.zip
date   mysite/test1.zip
date   mysite/test2.zip
date   mysite/test1.zip
date   mysite/test2.zip
size   mysite/test1.zip
exists mysite/test1.zip
size   mysite/test2.zip
exists mysite/test2.zip
```

as you can see, `exists()` calls are made after `date()` calls

src/BackupDestination/Backup.php
orottier: It's a precaution, I guess.

Since the `size()` call contains this snippet
```
        if (! $this->exists()) {
            return 0;
```

I'm lead to assume there have been occurences where `Backup` instances point to deleted files.

src/BackupDestination/Backup.php
freekmurze: I think the code is simpler if each method just caches its own result. I don't really care if we have one call extra that we could prevent. Could would be much easier to follow and performance will be better than it currently is.

Could you make this change?

orottier: sure, done

freekmurze:
Thanks!

orottier:
Great!

Thanks for sharing this package with the community

PR #952, opened 2019-09-03
Add support for Laravel 6.

mvdnbrk:


composer.json
```diff
@@ -19,21 +19,21 @@
     ],
     "require": {
         "php": "^7.2",
-        "illuminate/console": "~5.8.0",
-        "illuminate/contracts": "~5.8.0",
-        "illuminate/events": "~5.8.0",
-        "illuminate/filesystem": "~5.8.0",
-        "illuminate/notifications": "~5.8.0",
-        "illuminate/support": "~5.8.0",
-        "league/flysystem": "^1.0.49",
```
tabacitu: @mvdnbrk thank your for this PR. Something stands out to me here. 

I see you're required the ```league/flysystem-aws-s3-v3``` adapter instead of ```league/flysystem```. Was this intentional?

freekmurze:
Thank you for this, but I don't think there's a need to drop L5 support at this point.

Issue #960, opened 2019-09-25
Backup stops and remaining destinations fail if there is a notification error


joshhanley:
I'm experiencing an issue where remaining backup destinations don't run if sending a notification fails.

I have just setup laravel-backup to backup to local storage and dropbox. 
I have configured email notifications with mailgun.

This was all working fine, then mailgun disabled my account pending business verification.

What happens now is when backup is run, it backs up to local storage, tries to send a notifcation of success, which fails due to mailgun error, then the whole backup process fails and it doesn't copy the zip to dropbox. 

This is a concern if backups stop mid process due to an error on a service such as mailgun. I would expect it to attempt to copy the files to the remaining destinations regardless of notification status.

Is there a way for backup to continue even if notifications fail?

Thanks in advance!

freekmurze:
We'll handle this futher in #961

PR #961, opened 2019-09-25
Fix backup stopping on notification failure (#960)

joshhanley:
This ensures backups continue even if there is a notification failure (#960) and dumps the error to console. 

It doesn't distinguish which notification failed though. Not sure if that is an issue.

Hope this helps. Thank you for a great package!

### DIFF
--- a/src/Tasks/Backup/BackupJob.php
+++ b/src/Tasks/Backup/BackupJob.php
@@ -275,7 +275,11 @@ class BackupJob
     protected function sendNotification($notification)
     {
         if ($this->sendNotifications) {
-            event($notification);
+            try {
+                event($notification);
+            } catch (Exception $exception) {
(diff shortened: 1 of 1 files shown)


src/Tasks/Backup/BackupJob.php
```diff
@@ -275,7 +275,11 @@ protected function copyToBackupDestinations(string $path)
     protected function sendNotification($notification)
     {
         if ($this->sendNotifications) {
-            event($notification);
+            try {
```
freekmurze: Let's use the `rescue` helper instead of the try catch block: https://laravel.com/docs/master/helpers#method-rescue

freekmurze:
Thank you!

joshhanley:
Thanks for your guidance! I wasn't aware of the rescue function.

Issue #958, opened 2019-09-16
Set from address for notifications


Sinnbeck:
Any chance that it would be possible to set a custom sender address in the notifications?

https://laravel.com/docs/5.8/notifications#customizing-the-sender

PR #967, opened 2019-10-02
Custom From Address for Mail Notifications

damcclean:
There was an issue posted on the repo a few days ago, #958 which requests the ability to configure custom from addresses for mail notifications.

Laravel itself gives you the opportunity to change your from name and address. However, some users may want to use a different address for their backup emails.

So far I've managed to update the config file with the new values, update the documentation snippet and add the `->from` method to each of the mail notifications.

One thing I was not able to figure out was a way to test that the email is being sent from the email address we configured for the package, rather than the one set in Laravel's config. Let me know if you can think of a way to address this.


freekmurze:
Thanks!

sikhlana:
There's a huge issue with the pull request.

The call `config('backup.mail.from.address', 'mail.from.address')` should be `config('backup.mail.from.address', config('mail.from.address'))`. This goes for all the other `config()` calls.

I just got exceptions from all my projects that `mail.from.address` is not a valid RFC email address.

sikhlana:
Looks like the PR #973 fixes the issue. Please release an update as soon as possible.

PR #973, opened 2019-10-07
Fix email notifications not being sent

mixsuite:
There is an error in the config key and when the package tries to send the notification an exception is thrown: Cleanup failed because: Address in mailbox given [mail.from.address] does not comply with RFC 2822, 3.6.2..


freekmurze:
Thanks!

PR #974, opened 2019-10-08
Add disk getter function on Backup

patrickomeara:
This PR allows you to read `disk()` on `Backup` which helps when writing your own `CleanupStrategy`

```php
class MyCleanupStrategy extends DefaultStrategy
{
    public function deleteOldBackups(BackupCollection $backups)
    {
        // backups can only be from one disk per call, but all go through this function
        [$s3Backups, $dropboxBackups] = $backups->partition(
            function (Backup $backup) {
                return $backup->disk() === 's3';
            }
        );

        // s3 backups follow the DefaultStrategy
        parent::deleteOldBackups($s3Backups);

        // only keep the latest Dropbox backup
        $dropboxBackups->shift();
        $dropboxBackups->each->delete();
    }
}
```

### DIFF
--- a/src/BackupDestination/Backup.php
+++ b/src/BackupDestination/Backup.php
@@ -29,6 +29,11 @@ class Backup
         $this->path = $path;
     }
 
+    public function disk(): Filesystem
+    {
+        return $this->disk;
+    }
+
     public function path(): string
     {
         return $this->path;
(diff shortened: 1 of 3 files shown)


freekmurze:
Thanks!

SlyDeath:
Need to compare with instance, not string:
```php
[$s3, $local] = $backups->partition(fn(Backup $backup) => ($backup->disk() === \Storage::disk("s3")));
```

patrickomeara:
Ah you are correct @SlyDeath I have updated the example code for anyone stumbling upon this in the future!

👍🏻

PR #976, opened 2019-10-09
php 32 bit compatibility

schonhoff:
Tested these changes on my system and it worked with a 32bit php system. The notification and the backup:list command is working again with no failure.

It should be tested on a 64bit php system before!

### DIFF
--- a/src/BackupDestination/Backup.php
+++ b/src/BackupDestination/Backup.php
@@ -60,7 +60,7 @@ class Backup
     /**
      * Get the size in bytes.
      */
-    public function size(): int
+    public function size(): flaot
     {
         if ($this->size === null) {
             if (! $this->exists()) {
--- a/src/BackupDestination/Backup.php
+++ b/src/BackupDestination/Backup.php
@@ -60,7 +60,7 @@ class Backup
     /**
      * Get the size in bytes.
      */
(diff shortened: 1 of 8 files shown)


src/BackupDestination/Backup.php
```diff
@@ -60,7 +60,7 @@ public function date(): Carbon
     /**
      * Get the size in bytes.
      */
-    public function size(): int
+    public function size(): flaot
```
freekmurze: This type is wrong. Could you also add a test that hits this code, so we can't make a mistake in the future?

schonhoff: I'm on it. I will rewrite the unit tests so they will pass with the given float instead of integer.

schonhoff: New commit added where the typo is fixed and two tests where added for Backup and BackupCollection.

freekmurze:
Thanks!

src/BackupDestination/Backup.php
StanBarrows: Shouldn't be this "float"?

schonhoff: Yes but this is already fixed :)

PR #986, opened 2019-10-28
Monitor command should fail in CI if backups are considered unhealthy

MoogyG:
I would like to use Gitlab scheduled pipelines to monitor my application backups, but `php artisan backup:monitor` returns 0 whatever the result. With my suggestion, my pipeline will fail as excpected.


freekmurze:
Thanks!

PR #989, opened 2019-10-29
Fix non zero error code in monitoring command

MoogyG:
There was an error in my last pull request, `return 1` was in the `each` function so command was always returning 0.
I refactored the function ans I TESTED IT in a CI environment.
Sorry for wasting your time.


freekmurze:
Thanks!

PR #997, opened 2019-11-26
Fix doc requirements

anasbouzid:
According to [composer.json](https://github.com/spatie/laravel-backup/blob/master/composer.json) it should be PHP 7.2


freekmurze:
Thanks!

PR #1008, opened 2019-12-11
Empty directories are now being backed up

### LINKED ISSUE #859: empty folders will not be backed-up?
I have some empty folder in my project, it seems they are disappeared in the backup zip file. Is it possible to also backup empty folders?

theoxygen:
Fixes #859 with empty directories not being backed up


freekmurze:
Thank you!

PR #1012, opened 2019-12-14
MySQL Socket support

jimping:
When passing the "socket" option in "dump" section of the mysql part in database.php the host part will not be ignored. Now it does, so socket connection is possible.


freekmurze:
Could you add a test that would fail if it weren't for this PR?

nicoorfi:
I believe this will also be fixed here https://github.com/spatie/db-dumper/pull/145.

PR #1015, opened 2019-12-23
Add ability to define database connections using invokable class

daniel-de-wit:
My project is a multi-tenant / sharding setup that has a dynamic amount of database connections defined on the fly.

This PR allows me to define an invokable class in the config to determine the databases to backup.

For example:

config.php
```php
'database_generator' => \App\Support\Backups\DatabaseConnectionGenerator::class,
```

Generator
```php
<?php

namespace App\Support\Backups;

class DatabaseConnectionGenerator
{
    public function __invoke(): array
    {
         // return list of database connections
    }
}
```

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -70,6 +70,16 @@ return [
             'databases' => [
                 'mysql',
             ],
+
+            /*
+             * Alternatively, you may define an FQN for an invocable class that generates an array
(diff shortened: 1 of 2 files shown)


freekmurze:
To keep the package simple, I'm going to pass on this for now. But I'll revisit this PR when creating a new major version of the package.

PR #1024, opened 2020-01-24
Add doc entry about secure connections

sebwas:
I don't know if this is useful, but it took me quite a bit digging to find out how to fix this problem, so I thought it might be useful to others. Thus sharing.


freekmurze:
Hi, thanks for working around this. I'm not going to merge this in because I don't want to maintain this type of docs. If you would put this somewhere on a blog, I would gladly link to it.

PR #1025, opened 2020-01-27
MongoDB authentication uses correct connection name

artoroz:
When using a custom connection name the MongoDB does not use the authentication database due to a hardcoded reference


freekmurze:
Thanks!

PR #1046, opened 2020-03-12
Added support to set execution timeout

Caine85:
Added the timeout parameter to set the maximum execution time, this could be useful if the amount of data to backup is big.


freekmurze:
Thanks!

PR #1080, opened 2020-06-14
Avoid Errors on Unreadeable folder

ulefebvr:
This change avoid errors on unreadable folder.

When iterating on folder I had permission issue. Simply adding folder you don't have access to in `backup.source.files.exclude` doesn't avoid permission issue.

### DIFF
--- a/src/Tasks/Backup/FileSelection.php
+++ b/src/Tasks/Backup/FileSelection.php
@@ -68,6 +68,7 @@ class FileSelection
         }
 
         $finder = (new Finder())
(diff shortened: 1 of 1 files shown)


freekmurze:
I'm not certain I this is better if we were to add this by default. I think it's good that, when a directory can't be included in the backup, that it should try an error.

ulefebvr:
Ok, we can set the default behavior to throw an error.

Maybe add a config setting to activate//deactivate this?

freekmurze:
I'd be ok with add a config option for this.

PR #1081, opened 2020-06-15
Avoid Unreadable folders errors, can be set from config

ulefebvr:
I tried to respect the architecture and documentation, let me know if something does not respect the way of thinking of this package.

You can set `backup.source.files.ignore_unreadable_directories` in the config file.
If true, it will ignore the folders it doesn't have permission to access.
Otherwise, it will act normal and raise an error.


freekmurze:
Thanks!

PR #1113, opened 2020-09-08
add support for Laravel 8

tabacitu:
Good news! Looks like all we need to do to support Laravel 8 is bump the version requirements 🥳

I can't say I've tested this extensively, you guys know best, but... if the unit tests are running fine and my backups have run fine, I think we're good to go to support Laravel 8:

Screenshot 2020-09-08 at 10 12 30


tabacitu:
Not sure why some tests failed, ran fine in my env:
Screenshot 2020-09-08 at 10 42 11


I haven't jumped on the Github actions bandwagon yet, so I have no idea if adding "Laravel 8" to the matrix is what I was supposed to do, feedback here would be appreciated 😄

freekmurze:
I've taken care of this myself. Thanks!

PR #1121, opened 2020-09-21
Update required PHP version from 7.2 to 7.3 in Readme

Kovah:
By the way: wouldn't it be better to introduce a bump in the required PHP version as a major release? Was quite surprised that upgrading from 6.11.1 to 6.11.2 stopped working with PHP 7.2 suddenly.


freekmurze:
Thanks!

PR #1128, opened 2020-09-29
Added Portuguese Translations

hmrneves1:
Hello,

I have used this package in 2 projects in my course academy. Since the locale from the projects was "PT", the application doesn't get the right translation. Hope this is helpful to other people too.

Best Regards,
Helder


freekmurze:
Thanks!

PR #1136, opened 2020-10-19
use "local_secure" instead of "local" as default storage location

thyseus:
This avoids to unexpectedly save database and code into
a public space where it should not belong.
"local_secure" disk is configured in a laravel default
skeleton and should be available. The files will go into
the storage/ folder instead of the public/ folder now.

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -101,7 +101,7 @@ return [
              * The disk names on which the backups will be stored.
              */
             'disks' => [
-                'local',
+                'local_secure',
             ],
         ],
 


freekmurze:
I'm going to leave it like it is for now.

thyseus:
What about adding a notice to the installation instructions about the public available backups ? I see this as a minor security issue if people accidentally forget to set the path so a secure location.

freekmurze:
Feel free to send a PR to the installation instructions, explaining that it's recommend to create a disk named `backup` and use that one.

PR #1141, opened 2020-11-01
Test backup deletion

### LINKED ISSUE #1108: Inform about not working cleanup
I noticed that in [BackupDestination/Backup](https://github.com/spatie/laravel-backup/blob/85d6710ce77176ce926268a22825f9f70afb7aae/src/BackupDestination/Backup.php#L90) there is no test whether the file deletion was successfull or not.

When using [flysystem-webdav](https://github.com/thephpleague/flysystem-webdav/blob/b5c3c756e60cbd495173ce09c0a68a858803f4ce/src/WebDAVAdapter.php#L236) the delete method only returns true when the webdav server is responding with an 2xx status code.

In my opinion we should at least process the result of the delete method into an appropriate console output.

I am willing to submit a merge request if you are interested but I first wanted to discuss the topic.

akmet:
I added an extra test to check if the filesystem has deleted the backup.

This fixes #1108.

### DIFF
--- a/src/BackupDestination/Backup.php
+++ b/src/BackupDestination/Backup.php
@@ -89,9 +89,12 @@ class Backup
 
     public function delete()
     {
-        $this->disk->delete($this->path);
-        $this->exists = false;
+        if ($this->disk->delete($this->path) !== true) {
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #1149, opened 2020-11-19
Allow a CleanupStrategy to access its intended BackupDestination

JackWH:
My application backs up to two disks — local storage, and an S3 bucket. I want to keep just 7 days of backups locally, but a years' worth on S3.

I wrote a custom `CleanupStrategy` class, and set it as the strategy in the backup config file. I was intending to handle the logic for each disk in this custom class, but then realised the `CleanupJob` task only passes a `BackupCollection` to the strategy, with no (easy) way to determine where those backups are located.

This PR simply sets a `backupDestination` property on the abstract `CleanupStrategy` class, allowing custom strategies that extend this class to identify the intended backup destination and react accordingly.

This should be backwards-compatible with any classes implementing the abstract `CleanupStrategy` class.

Thanks for considering this change!

### DIFF
--- a/src/Tasks/Cleanup/CleanupJob.php
+++ b/src/Tasks/Cleanup/CleanupJob.php
@@ -39,7 +39,10 @@ class CleanupJob
 
                 consoleOutput()->info("Cleaning backups of {$backupDestination->backupName()} on disk {$backupDestination->diskName()}...");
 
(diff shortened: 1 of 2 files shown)


freekmurze:
Nice! Thanks!

PR #1153, opened 2020-11-27
Add temporary file paths to dumpDatabases function

wadimsewo:
Hello,

we have added a new db-dumper called "MyDumper" via the function DbDumperFactory::extend. It works multithreaded and creates one dump file per table. To recognize multiple files in the temporary folder, I extended your code to make it work. 

With kind regards,
Wadim


freekmurze:
Hi,

thanks for your work on this. I'm not going to merge this in, as I don't want to maintain it (and there were no tests provided)

PR #1154, opened 2020-12-01
Update README.md, docs and comment in config file

jrmajor:
- Update information about minimal required version of PHP and Laravel
- Update links to Laravel documentation to point to latest version
- Sending Slack notifications requires `laravel/slack-notification-channel`, which requires (in `composer.json`) `guzzlehttp/guzzle`. Comment in config file instructed to require both of them, and documentation didn't event mention `laravel/slack-notification-channel`. Fixed that!


jrmajor:
I've also found some random trailing whitespace in docs. It felt wrong to remove it, but so did feel leaving it as is. My editor auto-trimmed it, so I can commit these changes if you want.

freekmurze:
Thanks!

Issue #1065, opened 2020-04-30
Remove path in zip files.


letuananh1873:
Hi,

When I run backup:run --only-files, and open zip files, folder is
D_\path_xamp\my-product\all fiile

When i open zip file, i click multi folder to get root project.

Can i remove path file D:\path_xamp ?

Thanks

freekmurze:
This is currently not possible.

Issue #132, opened 2016-05-12
Stops at Determining files to backup...


snipe:
Hi all,

First, thanks so much for this great package! I'm running v3.7.0 on Laravel 5.2, and when I try to run `php artisan backup:run`, it seems to get stuck on determining the files to backup, so no backup file is produced.

```
snipe$ php artisan backup:run
Starting backup...
Dumping database snipeit_laravel...
Determining files to backup...
```

And that's it. Nothing in the error logs, I see a zip file created in `storage/laravel-backups/tmp`, but when I open that zip, it only has the SQL dump in there. The `storage/backups` destination directory is writable.

I don't see the old configuration of the destination directory, just the destination disk, but I'm not seeing the console feedback where it tells me how many files it's going to store. 

I only want to backup the config directory, and user uploaded files, all to the local filesystem. My config file is below:

```
<?php

return [

    'backup' => [

        /*
         * The name of this application. You can use this name to monitor
         * the backups.
         */
        'name' => env('APP_URL'),

        'source' => [

            'files' => [

                /*
                 * The list of directories that should be part of the backup. You can
                 * specify individual files as well.
                 */
                'include' => [
                    base_path('public/uploads'),
                    base_path('config'),
                    base_path('storage/private_uploads'),
                ],

                /*
                 * These directories will be excluded from the backup.
                 * You can specify individual files as well.
                 */
                'exclude' => [
                    base_path('vendor'),
                    base_path('node_modules'),
                ],

                /*
                 * Determines if symlinks should be followed.
                 */
                'followLinks' => false,
            ],

            /*
             * The names of the connections to the databases that should be part of the backup.
             * Currently only MySQL- and PostgreSQL-databases are supported.
             */
            'databases' => [
                'mysql',
            ],
        ],

        'destination' => [

            /*
             * The disk names on which the backups will be stored.
             */
            'disks' => [
                'local',
            ],
        ],
    ],

    'cleanup' => [
        /*
         * The strategy that will be used to cleanup old backups.
         * The youngest backup will never be deleted.
         */
        'strategy' => \Spatie\Backup\Tasks\Cleanup\Strategies\DefaultStrategy::class,

        'defaultStrategy' => [

            /*
             * The amount of days that all daily backups must be kept.
             */
            'keepAllBackupsForDays' => 7,

            /*
             * The amount of days that all daily backups must be kept.
             */
            'keepDailyBackupsForDays' => 16,

            /*
[...]

PR #1158, opened 2020-12-08
add option to add files to zip-file using relative path

marnickmenting:
As mentioned here: https://github.com/spatie/laravel-backup/issues/132#issuecomment-219143235
And here: https://github.com/spatie/laravel-backup/issues/1065

This pull request adds an option to enter a relative path that is removed from the full path of the files in the resulting ZIP-file.

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -40,6 +40,13 @@ return [
                  * Determines if it should avoid unreadable folders.
                  */
                 'ignore_unreadable_directories' => false,
+
+                /*
+                 * This path is used to make directories in resulting zip-file relative
+                 * Set to false to include complete absolute path
+                 * Example: base_path()
+                 */
+                'relative_path' => false,
             ],
 
             /*
--- a/docs/installation-and-setup.md
+++ b/docs/installation-and-setup.md
@@ -57,6 +57,13 @@ return [
                  * Determines if symlinks should be followed.
                  */
                 'follow_links' => false,
+
+                /*
+                 * This path is used to make directories in resulting zip-file relative
+                 * Set to false to include complete absolute path
+                 * Example: base_path()
+                 */
+                'relative_path' => false,
             ],
 
             /*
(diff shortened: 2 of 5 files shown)


marnickmenting:
It seems some tests failed because of connection issues with Microsoft servers, but I don't know how to run them again.

config/backup.php
```diff
@@ -40,6 +40,13 @@
                  * Determines if it should avoid unreadable folders.
                  */
                 'ignore_unreadable_directories' => false,
+
+                /*
+                 * This path is used to make directories in resulting zip-file relative
+                 * Set to false to include complete absolute path
+                 * Example: base_path()
+                 */
+                'relative_path' => false,
```
freekmurze, line 49: Most of our config options start with a verb. Rename this to `use_relative_path` for consistency/

freekmurze (changes requested):
Great feature! Might change this to `true` in the next major version of the package.

config/backup.php
marnickmenting: Actually, it works by providing a path. So if I set 'relative_path' to '/Users/marnick/' It will remove that part from the path in the zip-file. Not sure if that is clear enough explained though.

I might change the option to a boolean, and just remove the base_path() from all included files if set to true. But for my use case I needed a more flexible option (dealing with tenancy).

freekmurze:
Could you take a look at the failing tests?

marnickmenting:
> Could you take a look at the failing tests?

Yes, like I said here https://github.com/spatie/laravel-backup/pull/1158#issuecomment-740553314, it seems it had connection problems with Microsoft when running the tests, and therefore failed. But I don't know how to trigger them again. Should I do an empty commit?

freekmurze:
> But I don't know how to trigger them again.
I think only the repo owner can do it. Just rerun them, all is fine 👍

freekmurze:
Thanks!

PR #1161, opened 2020-12-14
V7

freekmurze:
- suffix classes
- code cleanup
- drop support for PHP 7
- drop support for anything below Laravel 8


MatanYadaev:
@freekmurze Amazing 👏  Please, keep pushing the PHP community forward!

marnickmenting:
Isn't it a bit fast to drop support for PHP7.4? I believe Laravel 8 still supports PHP7.4. I have some other dependencies that need an update to support PHP8.

Issue #1126, opened 2020-09-29
Backup destination hiding important error message


antonioribeiro:
If you don't configure the destination disk properly, it will be very hard to understand what's happening because this message:

**"Endpoints must be full URIs and include a scheme and host"**

Is hidden by the exception handler:

``` php
public static function create(string $diskName, string $backupName): self
{
    try {
        $disk = app(Factory::class)->disk($diskName);

        return new static($disk, $backupName, $diskName);
    } catch (Exception $exception) {
        $backupDestination = new static(null, $backupName, $diskName);

        $backupDestination->connectionError = $exception;

        return $backupDestination;
    }
}
```

And all you get is 

```
Copying zip failed because: There is no disk set for the backup named `Backup-Name`..
```

daisuke85a:
@freekmurze 
Could you assign me this issue?
I want to challenge first issue.

freekmurze:
@daisuke85a feel free to send in a PR to fix this.

freekmurze:
Fixed in the latest release

PR #1162, opened 2020-12-15
Fix to show Backup destination hiding important error message

masterfermin02:
Issue: #1126
This pull request is to show Backup destination hiding important error messages.


freekmurze:
Thanks!

PR #1167, opened 2020-12-21
Add missing visibility keyword

rodrigopedra:
This PR adds missing visibility keywords to promoted properties.

### DIFF
--- a/src/Events/BackupHasFailed.php
+++ b/src/Events/BackupHasFailed.php
@@ -8,7 +8,7 @@ use Spatie\Backup\BackupDestination\BackupDestination;
 class BackupHasFailed
 {
     public function __construct(
-        Exception $exception,
-        ?BackupDestination $backupDestination = null,
+        public Exception $exception,
+        public ?BackupDestination $backupDestination = null,
     ) {}
 }
--- a/src/Events/CleanupHasFailed.php
+++ b/src/Events/CleanupHasFailed.php
@@ -8,7 +8,7 @@ use Spatie\Backup\BackupDestination\BackupDestination;
 class CleanupHasFailed
 {
     public function __construct(
-        Exception $exception,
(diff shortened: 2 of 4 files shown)


freekmurze:
Thanks!

rodrigopedra:
Hi @freekmurze 

Thanks for merging it. When I first found a missing visibility keyword I forked the package and checked every constructor for missing visibility keywords.

While checking those I thought on converting other constructors to leverage promoted properties, those would be the ones with some initialization done in the constructor.

Then I stepped back as I didn't see any such mixed usage (promoted properties + regular property initialization), and thought it could be a design decision from your team. Also I didn't want to make this PR big, as the ones I sent actually fixes potential bugs as those properties won't be really set not available if they were not promoted (with the exception of the `EventHandler`, where it only had one property and I decided to convert it to a promoted property).

As an example (I don't know if I am being quite clear)

https://github.com/spatie/laravel-backup/blob/585e1e7e09a1eca7d2d000d2192c6bfe4404f6ba/src/Tasks/Cleanup/CleanupJob.php#L14-L30

In this constructor both `$backupDestinations` and `$strategy` could be converted to promoted properties.

If you want me to review those I can send another PR for those cases. 

If you prefer to not mixed promoted properties and regular constructor assignments I totally get it. Actually I didn't decided yet on this matter for the projects I am migrating to PHP 8.

Thanks again and have a nice day =)

freekmurze:
We only use promoted properties if every property can be a promoted one: https://spatie.be/guidelines/laravel-php#constructor-property-promotion

 In that `CleanUpJob` we don't used promoted properties because `sendNotifications` has that negation going on in the assignment.

I think it's ok to leave it like this.

rodrigopedra:
Thanks for the quick response. I think I'll adopt the same guideline on my projects.

PR #1202, opened 2021-01-11
Small improvement suggestion for the backup times

ndeblauw:
It's really not that important, but it could make a difference for someone (given your large user base), so I decided to suggest it. No explanation needed, nor bad feelings, if you decide just to discard it :-)


freekmurze:
Thanks!

Issue #1225, opened 2021-01-28
Error notification is not sent when backup disk is not reachable


acabutto:
When a backup disk (e.g. over ftp) is not reachable for whatever reason (e.g. wrong credentials, network failure, etc...) the corresponding error notification is not sent because the notification itself tries to retrieve information from the backup disk for reporting purposes.

I am currently working on a pull request, should come shortly.

PHP: 7.4
Laravel: 8.0
Laravel-Backup: Latest

PR #1227, opened 2021-01-28
Fix for #1225

acabutto:
Already merged on master but useful for php 7.4 users that are still using laravel-backup 6.x


freekmurze:
Thanks!

Issue #1209, opened 2021-01-17
Backup Events not fire when useing --disable-notifications flag


AltiumWilliam:
Hi
I'm using **BackupZipWasCreated** Event to encrypt backup files, I found that when I use **_--disable-notifications_** flag, Events not fire but when I clear the flag, all events fire normally.
I register Event in **EventServiceProvider** and every thing is ok, only flag make the event not firing.

Laravel version: 5.8.38
PHP version: 7.4
Laravel spatie-backup version: 6.12.0

freekmurze:
I'd accept a PR that fixes this.

FelipeVeiga:
same problem here

acabutto:
> Hi
> I'm using **BackupZipWasCreated** Event to encrypt backup files, I found that when I use **_--disable-notifications_** flag, Events not fire but when I clear the flag, all events fire normally.
> I register Event in **EventServiceProvider** and every thing is ok, only flag make the event not firing.
> 
> Laravel version: 5.8.38
> PHP version: 7.4
> Laravel spatie-backup version: 6.12.0

I may be wrong but after having a quick look at the code I think this is not a bug but a by design behavior instead. The misunderstanding here could be related to naming of the switch **_--disable-notifications_**. It should be named **_--disable-event-firing_** because that is exactly what it does:

```
[...]
$this->sendNotification(new BackupZipWasCreated($pathToZip));
[...]

protected function sendNotification($notification)
    {
        if ($this->sendNotifications) {
            rescue(function () use ($notification) {
                event($notification);
            }, function () {
                consoleOutput()->error('Sending notification failed');
            });
        }
    }
```
Whether to notify the user or not after an event is something that should be handled in your project possibly in `config/backups.php`

```
    'notifications'   => [
        'notifications' => [
            BackupHasFailed::class         => ['mail'],
            UnhealthyBackupWasFound::class => ['mail'],
            CleanupHasFailed::class        => ['mail'],
            BackupWasSuccessful::class     => ['mail'],
            HealthyBackupWasFound::class   => ['mail'],
            CleanupWasSuccessful::class    => ['mail'],
        ],
```
Am I right?

freekmurze:
It seems you are right, and I'll consider improving this name/behaviour in the next major version.

AltiumWilliam:
I think, It must be separated, disable events and disable notifications. User must to able to select whether to send notification (broadcast, email,...) or events. Sometimes I need Events and Sometime I need Notifications.
[...]

PR #1231, opened 2021-02-05
Disable notifications refactoring

acabutto:
As per #1209 issue discussion I have put in place a little notifications / events disable feature refactoring.

The `--disable-notifications` flag acts exactly as its name says, so no notification will be sent.
A new `--disable-events` flag has been introduced to actually disable events firing.

Notification cannot be sent when events firing is disabled.

Tests has been added to check this new behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ composer.lock
 vendor
 .phpunit.result.cache
 .php_cs.cache
+.idea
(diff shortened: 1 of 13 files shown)


freekmurze:
As this PR contains breaking changes, I'm going to close it for now. I'll revisit it when creating a new major version.

acabutto:
All right. Let me know if you need any change or support to merge it.

PR #1236, opened 2021-02-17
use LOG_SLACK_WEBHOOK_URL as default for slack notifications

johanrosenson:
This is a minor change that defaults to use the already present LOG_SLACK_WEBHOOK_URL used by Laravel for the Slack log channel, this makes it easier to configure the slack notifications, especially if you already are using the slack log channel and wish to use the same webhook for it.

This should be a pretty safe default since slack is not enabled on any of notifications anyway, so no unwanted notifications should be sent.

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -160,7 +160,7 @@ return [
         ],
 
         'slack' => [
-            'webhook_url' => '',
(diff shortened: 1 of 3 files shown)


freekmurze:
Thanks for the PR, I'm going to pass on it for now.

PR #1239, opened 2021-02-23
Allow to encrypt backup archives

Gummibeer:
This PR is a community contribution - I'm only the messenger.

@olssonm has created a cool add-on package that adds this logic.
https://github.com/olssonm/laravel-backup-shield

@skollro has created a blog post on how to add this logic to your own app.
https://simonkollross.de/posts/creating-encrypted-backups-of-laravel-apps

@mstaack has implemented this logic in our own apps which was the third source for this PR.

As this feature is an important security improvement and all users should benefit I'm proposing these changes to the main package.

If/after this PR is accepted I will, if wanted, check how to migrate this to the v7 branch. But as we use PHP7.4 we stick with v6 so I wanted to add this feature to the v6 first as upgrading seems easier than downgrading.


Gummibeer:
@freekmurze if you have some time it would be great if you can review this PR. I think that a lot of people can benefit from it to have more secure backups (by default).

docs/installation-and-setup.md
```diff
@@ -58,12 +58,17 @@ return [
                  */
                 'follow_links' => false,
 
+                /*
+                 * Determines if it should avoid unreadable folders.
```
freekmurze, line 62: It might be good the explain what would happen if this is set to `true` and an unreadable directory is processed. Will an exception be thrown?

freekmurze:
Awesome work, thank you! I'll already merge this in a tag it.

I would highly appreciate a PR that brings this feature to the latest version of the package as well 👍

docs/installation-and-setup.md
Gummibeer: I've just copied the real config to here as it was out of sync. 😅

Gummibeer:
Will check how to migrate.
Should I make a PR from v6 to master branch or do it completely new? 🤔
I'm not sure what's the best approach here with the version branches. But I think I should do it "new" for the master as in v6 are other changes that shouldn't move up!?

freekmurze:
I think it might be easiest to do a new PR for v7 👍

config/backup.php
```diff
@@ -103,8 +103,8 @@
          * If not specified, the file extension will be .archive for MongoDB and .sql for all other databases
          * The file extension should be specified without a leading .
          */
-        'database_dump_file_extension' => '',        
-        
+        'database_dump_file_extension' => '',
```
Gummibeer, line 106: PHPstorm auto-format

config/backup.php
```diff
@@ -124,6 +124,18 @@
...
+         * The password to be used for archive encryption.
+         * Set to `null` to disable encryption.
+         */
+        'password' => env('BACKUP_ARCHIVE_PASSWORD'),
+
+        /*
+         * The encryption algorithm to be used for archive encryption.
+         * You can set it to `null` or `false` to disable encryption.
+         */
+        'encryption' => \ZipArchive::EM_AES_256,
```
Gummibeer, line 138: real added config

docs/installation-and-setup.md
```diff
@@ -133,6 +146,18 @@ return [
...
+         * The password to be used for archive encryption.
+         * Set to `null` to disable encryption.
+         */
+        'password' => env('BACKUP_ARCHIVE_PASSWORD'),
+
+        /*
+         * The encryption algorithm to be used for archive encryption.
+         * You can set it to `null` or `false` to disable encryption.
+         */
+        'encryption' => \ZipArchive::EM_AES_256,
```
Gummibeer, line 160: All changes except of this are only the copied real config as this document was out of sync.

composer.json
```diff
@@ -19,6 +19,7 @@
     ],
     "require": {
         "php": "^7.3|^8.0",
+        "ext-zip": "^1.14.0",
```
Gummibeer, line 22: Required because of https://www.php.net/manual/de/ziparchive.setencryptionindex.php

PR #1246, opened 2021-03-04
Encrypt backup archives v7

### LINKED ISSUE #1245: Port "Allow to encrypt backup archives" PR to v7
The PR #1239 was submitted for v6 and needs to be ported to v7.

I'm currently working on this issue.

This issue is to help keep track of who's working on the port.

@freekmurze @Gummibeer

patinthehat:
This PR submits a port of #1239 for `v7` _(the original PR adds support for encrypting backup archives for for `v6`)_.  All changes from #1239 have been updated to work with `v7`.

As with the original PR, there are also some minor updates to the `installation-and-setup` docs.

Resolves #1245.

@freekmurze @Gummibeer


freekmurze:
Looks good! Have you also tested this manually, just to be sure that you can decrypt the zips with the right password?

patinthehat:
> Looks good! Have you also tested this manually, just to be sure that you can decrypt the zips with the right password?

Yes. I was able to decrypt the zips with the correct password without any issues.

Gummibeer:
As you've asked: for me as Mac user the only problem was that the default Archive/Zip application on macOS isn't able to extract encrypted archives.
So I had to use another application (Unarchiver) but it was the same for all archives, so seems like a problem with the app and not the packages logic.

freekmurze:
Thank you very much!

drbyte:
I tested it on the v6 branch and BetterZip (Mac) can read the archives fine after prompting for the password. 
Clarification: zips created and encrypted on ubuntu server via this package; and I unzipped on my Mac using BetterZip

Gummibeer:
Yeah, it's only the default program used by right-click "compress" on a folder too.
It just fails with an error and says that the ZIP is corrupted - which isn't the case! The program is simply unable to open encrypted archives. 😂

Possibly worth a note on the docs => will send PRs.

drbyte:
> default Archive/Zip application on macOS isn't able to extract encrypted archives

I confirm the same symptom using the default macOS Archive application. (In my case macOS 10.14/Mojave.)

Issue #1244, opened 2021-03-03
Undefined class constant 'EM_AES_256' when installing laravel-backup


eduardoarandah:
# What happened

when I do `composer require spatie/laravel-backup` I get `Undefined class constant 'EM_AES_256'`

```
Generating optimized autoload files
ocramius/package-versions: Generating version class...
ocramius/package-versions: ...done generating version class
> Illuminate\Foundation\ComposerScripts::postAutoloadDump
> @php artisan package:discover --ansi

   Symfony\Component\Debug\Exception\FatalThrowableError  : Undefined class constant 'EM_AES_256'

  at /home/donors/tmp/donors/vendor/spatie/laravel-backup/config/backup.php:138
    134|         /*
    135|          * The encryption algorithm to be used for archive encryption.
    136|          * You can set it to `null` or `false` to disable encryption.
    137|          */
  > 138|         'encryption' => \ZipArchive::EM_AES_256,
    139|     ],
    140|
    141|     /*
    142|      * You can get notified when specific events occur. Out of the box you can use 'mail' and 'slack'.

  Exception trace:

  1   require()
      /home/donors/tmp/donors/vendor/laravel/framework/src/Illuminate/Support/ServiceProvider.php:64

  2   Illuminate\Support\ServiceProvider::mergeConfigFrom()
      /home/donors/tmp/donors/vendor/spatie/laravel-backup/src/BackupServiceProvider.php:38

  Please use the argument -v to see more details.
Script @php artisan package:discover --ansi handling the post-autoload-dump event returned with error code 1
```

# The environment

Ubuntu Server 
PHP 7.4.15
Composer version 2.0.9

# My guess

I guess it's something wrong about zip extension.

However it seems to be present:

```

php -i | grep zip
/etc/php74rc/conf.d/zip.ini
Registered PHP Streams => https, ftps, compress.zlib, php, file, glob, data, http, ftp, compress.bzip2, phar, zip
Registered Stream Filters => zlib.*, string.rot13, string.toupper, string.tolower, string.strip_tags, convert.*, consumed, dechunk, bzip2.*, convert.iconv.*
Stream Wrapper support => compress.bzip2://
Stream Filter support => bzip2.decompress, bzip2.compress
gzip compression => enabled
bzip2 compression => enabled
zip
Libzip version => 1.1.2

cat /etc/php74rc/conf.d/zip.ini
extension=zip.so

php -m | grep zip
zip

```

Now, check this out:

class exists:

```
[...]

PR #1254, opened 2021-03-10
fix: zip module undefined constant in config

eduardoarandah:
constant \ZipArchive::EM_AES_256 isn't always present in zip module

here:
https://www.php.net/manual/en/zip.constants.php

says:
AES 256 encryption. Available as of PHP 7.2.0 and PECL zip 1.14.0, respectively, if built against libzip ≥ 1.2.0.

this can break composer dump-autoload in ran in a system with zip module
built with older libzip

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -135,7 +135,9 @@ return [
          * The encryption algorithm to be used for archive encryption.
          * You can set it to `null` or `false` to disable encryption.
          */
-        'encryption' => \ZipArchive::EM_AES_256,
+        "encryption" => defined("\ZipArchive::EM_AES_256")
+            ? \ZipArchive::EM_AES_256
+            : null,
     ],
 
     /*


eduardoarandah:
This fix simply checks constant `\ZipArchive::EM_AES_256` exists before assigning it. 

For it to exist, php-zip extension has to be compiled against libzip >= 1.2.0 

In discussion, some common configurations affected: 

- cPanel v94.0.2 (the latest march 2021) under CentOS 7.9
- php:7.4-fpm-alpine Docker container

Before this fix, `composer update` or `composer install` breaks before creating autoload making the app unusable. 

Fixes issue: #1244 

Discussion: https://github.com/spatie/laravel-backup/discussions/1247

posglobalcadev:
Great to see a fix for this, I've been having the same problem with v6.15.0

freekmurze:
I don't like to have conditions in the config file, so I've solved this in a (imho) slightly cleaner way: https://github.com/spatie/laravel-backup/commit/71f6378fc48940c6a5898df24113dee65ddfee00

Thanks for your PR nonetheless 👍

PR #1267, opened 2021-03-16
fix: zip module undefined constant in config #1254 fo version 6

nunomazer:
Apply changes of version 7 to version 6 of this package


freekmurze:
Thanks!

Issue #1309, opened 2021-05-07
Artisan::call() throws a SIGINT undefined error


jacob-hyde:
Using the Artisan facade to make a call to a command with this package installed throws an error: Undefined constant \"Spatie\\Backup\\Commands\\SIGINT\"

Is there any known workarounds for this?

Related issue: https://github.com/spatie/laravel-backup/issues/1288

@SuddenlyRust comment:

I have encountered the same problem on my forge machine (linux). It happens when I use the `Artisan`-Facade to call any command.

Reproduction:

1. Add a route to your `web.php`-file:

```php
use App\Console\Commands\MyCommand;
use Illuminate\Support\Facades\Artisan;

Route::get('/reproduction', function () {
    Artisan::call(MyCommand::class);

    return '👀';
});
```

2. Visit `/reproduction` with your browser.

3. Following exception is thrown:
`Undefined constant "Spatie\Backup\Commands\SIGINT"`

---

OS: Ubuntu 20.04.2 LTS
PHP: v8.0.3
Webserver: nginx/1.18.0
Laravel: v8.37.0
spatie/laravel-backup: v7.5.1

_Originally posted by @SuddenlyRust in https://github.com/spatie/laravel-backup/issues/1288#issuecomment-821182809_

medvinator:
Same error on Ubuntu

JuanDMeGon:
Can confirm. It happens when calling any command using the Artisan facade (for some reason).

PR #1311, opened 2021-05-18
Fix issue #1309

mafeijam:
this is to fix issue #1309 when executing command from the web instead of CLI

### DIFF
--- a/src/Commands/BaseCommand.php
+++ b/src/Commands/BaseCommand.php
@@ -13,7 +13,7 @@ abstract class BaseCommand extends SignalAwareCommand
 
     public function __construct()
     {
-        if (PHP_OS_FAMILY !== 'Windows' && defined('SIGINT')) {
+        if (PHP_OS_FAMILY !== 'Windows' && $this->runningInConsole() && defined('SIGINT')) {
             $this->handlesSignals[] = SIGINT;
         }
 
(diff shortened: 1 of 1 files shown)


freekmurze:
Could we improve on this by only doing the `defined` check?

mafeijam:
just borrow the idea from Laravel Application class - `runningInConsole` method

maybe we can check for `in_array(php_sapi_name(), ['cli', 'phpdbg'])` to ensure the command is running in console instead of dong `defined` check.

karticbedi:
@freekmurze any ETA when you will consider this commit ? we are waiting to push our code to production because of this error.

freekmurze:
Checking on `in_array(php_sapi_name(), ['cli', 'phpdbg'])` and `defined('SIGINT')` seems like the way to go here.

src/Commands/BaseCommand.php
```diff
@@ -26,4 +26,9 @@ public function run(InputInterface $input, OutputInterface $output): int
 
         return parent::run($input, $output);
     }
+    
+    protected functiin runningInConsole(): bool
```
cretueusebiu, line 30: Typo here `functiin` -> `function`

freekmurze:
Thanks!

PR #1320, opened 2021-06-01
Add translations for backup destination properties

marcreichel:
This PR provides translations for the backup destination properties. Especially the keys.


freekmurze:
I'm going to pass on this, as I don't want to maintain this. I think for notifications it makes sense to have translations as maybe non-developers might see them. I'm going to assume that people that execute the artisan command, can speak English.

marcreichel:
@freekmurze But this PR was about the notifications. Not the artisan command 🤔 As I edited the `BaseNotification`.
So the emails (e.g.) are containing these (english) strings while the rest (subject and message) is translated (in german in my case). Or do I get something wrong?

image

freekmurze:
Right! In that case, I'll have it!

Thanks for your work on this!

marcreichel:
I'm happy I've changed your mind 😄
Thanks for merging 😊

PR #1328, opened 2021-06-05
Fix: BackupDestination does not check write result

### LINKED ISSUE #1326: BackupDestination does not check write result

rbrishabh:
Fixes #1326 

Hi @freekmurze, how does this look?
I'm still learning PHP so if this needs any changes, please let me know!

Thank you


freekmurze:
Thanks @rbrishabh

What I learn from the issue is that the problem is that the code doesn't throw an exception, but that the result is empty. We should check for that, and also write a test to verify that the problem is solved.

Closing this for now.

PR #1347, opened 2021-07-09
Detect write failures

### LINKED ISSUE #1326: BackupDestination does not check write result

coatesap:
Fixes #1326 

This is an important change, as currently laravel-backup reports all misconfigured S3 backups as successful!

This change only applies to laravel-backup v6, which uses flysystem v1 (laravel-backup v7 uses flysystem v2 which has a different interface).


freekmurze:
Thanks!

PR #1348, opened 2021-07-09
adds: exception to throw when flysystem returns false

### LINKED ISSUE #1326: BackupDestination does not check write result

ajuchacko91:
fix #1326 

I have done research on this and I would like to note it here.
league/flysystem - v1: Methods in League\Flysystem\FilesystemInterface doesn't throw exceptions instead returns true, false, resource, array. In flysystem v2 there are [exceptions which will be thrown](https://flysystem.thephpleague.com/v2/docs/usage/exception-handling/) instead of returning false.

In this package only Backup@delete checks for its result for rest of the methods(writeStream, readStream) I've added checking the result and throwing exception.

Even though I tried to add tests for these cases, I find it difficult to achieve

### DIFF
--- a/src/BackupDestination/Backup.php
+++ b/src/BackupDestination/Backup.php
@@ -6,6 +6,7 @@ use Carbon\Carbon;
 use Illuminate\Contracts\Filesystem\Filesystem;
 use InvalidArgumentException;
 use Spatie\Backup\Tasks\Backup\BackupJob;
+use Spatie\Backup\Exceptions\InvalidBackupFile;
 
 class Backup
 {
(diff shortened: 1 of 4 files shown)


freekmurze:
Is this one ready for review?

ajuchacko91:
sort of. 

I am trying to come up with a meaningful test case to document these change, 
I would like to give it a shot once more. need 2 more days.

I trying to figure out how to make a two tests where it should ```throw an exception while reading/stream or writing stream```.
Any thoughts on this would be helpful

thank you

erikn69:
`$this->expectException(UnableToReadFile::class);`
Maybe it´s that
https://github.com/thephpleague/flysystem/blob/a3c694de9f7e844b76f9d1b61296ebf6e8d89d74/src/AdapterTestUtilities/FilesystemAdapterTestCase.php#L585-L592

ajuchacko91:
Hi @erikn69,

that's not what I meant but thank you.

ajuchacko91:
This PR is ready for review @freekmurze

freekmurze (changes requested):
Looks good, only one minor nitpick

ajuchacko91:
🤞

freekmurze:
Thanks!

PR #1352, opened 2021-07-19
Support when no signal's extension

erikn69:
Use signals only when ext-pcntl is available

It's a workaround after [Simfony/Console 681a074](https://github.com/symfony/console/commit/681a074054bc6fc00dfeb5f862f266847d9a8465), it is gonna be on the next release, This PR allows avoid signals on windows and linux without ext-pcntl, so dependency can be removed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,6 @@
     ],
     "require": {
         "php": "^8.0",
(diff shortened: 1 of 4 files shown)


freekmurze:
Very nice, thank you!

Issue #1312, opened 2021-05-21
--disable-notifications also disable encryption


fab2s:
Hello, 

When backups are run with `--disable-notifications` options, the resulting zip file is never encrypted.

It looks like this is coming from the fact encryption occurs as an event wrapped in the send notification method in `BackupJob.php`:
```php
    protected function sendNotification($notification): void
    {
        if ($this->sendNotifications) {
            rescue(
                fn () => event($notification),
                fn () => consoleOutput()->error('Sending notification failed')
            );
        }
    }
````

Thus never throwing the `BackupZipWasCreated `event triggering encryption:
```php
    $this->sendNotification(new BackupZipWasCreated($pathToZip));
````

Hope this helps, and keep up with the good work 👍

freekmurze:
Feel free to send a PR to fix this 👍

erikn69:
@freekmurze must be closed
https://github.com/spatie/laravel-backup/blob/11eb9f82bc0bd25ec69f5c169dde07290d913ce8/tests/Commands/BackupCommandTest.php#L380-L395
https://github.com/spatie/laravel-backup/blob/11eb9f82bc0bd25ec69f5c169dde07290d913ce8/src/Tasks/Backup/BackupJob.php#L230-L234

PR #1356, opened 2021-07-28
Fix 1312 Extracted encryption process out of sendNotification

aFluxx:
Hello there,

This pull request is my solution for issue #1312. I pulled the encryption process from the sendNotifcation method, so the event is also fired when the command is run with the --disable-notifications parameter.

This is my first "real" contribution to open source, so any advice is welcome!

### DIFF
--- a/src/Tasks/Backup/BackupJob.php
+++ b/src/Tasks/Backup/BackupJob.php
@@ -227,7 +227,9 @@ class BackupJob
 
         consoleOutput()->info("Created zip containing {$zip->count()} files and directories. Size is {$zip->humanReadableSize()}");
 
-        $this->sendNotification(new BackupZipWasCreated($pathToZip));
+        $this->encryptBackup($pathToZip);
+
+        $this->sendNotification($pathToZip);
 
         return $pathToZip;
     }
@@ -292,6 +294,14 @@ class BackupJob
                     $this->sendNotification(new BackupHasFailed($exception, $backupDestination ?? null));
                 }
             });
+    }
+
+    protected function encryptBackup($pathToZip): void
+    {
+        rescue(
+            fn () => event(new BackupZipWasCreated($pathToZip)),
+            consoleOutput()->info('Encrypting backup successful...')
+        );
     }
 
     protected function sendNotification($notification): void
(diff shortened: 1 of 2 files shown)


src/Tasks/Backup/BackupJob.php
```diff
@@ -227,7 +227,9 @@ protected function createZipContainingEveryFileInManifest(Manifest $manifest): s
 
         consoleOutput()->info("Created zip containing {$zip->count()} files and directories. Size is {$zip->humanReadableSize()}");
 
-        $this->sendNotification(new BackupZipWasCreated($pathToZip));
+        $this->encryptBackup($pathToZip);
+
+        $this->sendNotification($pathToZip);
```
erikn69, line 232: Why `$this->sendNotification($pathToZip);`?

aFluxx: You are right, the flow I worked out seems redundant. I try solving this later and come up with a better solutioN.

fredsal: Just delete that line

aFluxx: I will do that today. Sorry for the delay, I've been busy and I forgot about this!

src/Tasks/Backup/BackupJob.php
```diff
@@ -227,7 +227,7 @@ protected function createZipContainingEveryFileInManifest(Manifest $manifest): s
 
         consoleOutput()->info("Created zip containing {$zip->count()} files and directories. Size is {$zip->humanReadableSize()}");
 
-        $this->sendNotification(new BackupZipWasCreated($pathToZip));
+        $this->encryptBackup($pathToZip);
```
freekmurze, line 230: I don't understand why the function is named this way. It doesn't encrypt the backup, only sends a notification.

erikn69: https://github.com/spatie/laravel-backup/blob/daaf928f654731a8e9e9e1607ddd9c30a9a57d20/config/backup.php#L148-L149
Also if you create a `BackupZipWasCreatedNotification` class and add it to the config file, the notification will still be sent because the event `BackupZipWasCreated` is triggered, that makes `--disable-notifications` useless

aFluxx:
Closing this. @erikn69 fixed it in #1361

PR #1363, opened 2021-08-09
[v7] Fix exclude files when there are same start but diferent endings

### LINKED ISSUE #1357: File exclusion logic excludes files with same prefix
If you have a basic configuration like this which should backup everything but excludes the `.env` file (in reality more needs to be excluded) the `.env` file is correctly excluded but the `.env.example` too. So everything with the same prefix is excluded:

```php
'files' => [
    /*
     */
    'include' => [
        base_path(),
    ],

    /*
     *
     */
     'exclude' => [
        base_path('.env'), // .env.example
        base_path('.git'), // .git folder is excluded but .gitignore and .gitattributes too
    ],
```

erikn69:
New solution for 
Closes #1357

### DIFF
--- a/src/Tasks/Backup/FileSelection.php
+++ b/src/Tasks/Backup/FileSelection.php
@@ -103,8 +103,16 @@ class FileSelection
 
     protected function shouldExclude(string $path): bool
     {
+        $path = realpath($path);
+        if (is_dir($path)) {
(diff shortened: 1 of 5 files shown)


fredsal:
@freekmurze i wonder how long it would take to get this one "officially" v7 release?

PR #1366, opened 2021-08-10
[V6] Fix bugs

### LINKED ISSUE #1298: relative_path option includes all folders using full path
Using 7.5.2 of spatie/laravel-backup, with a vanilla Laravel 8 install.

Only thing I've changed in the backup.php config:
```
'relative_path' => base_path()
```
It does backup all the files using a [relative path](https://github.com/spatie/laravel-backup/pull/1158).
But: it also creates all the folders (without the files) with full path: /Users/path_to_source_code

E.g. when I `unzip 2021-04-26-14-45-37.zip`:
```
# files have been extracted, but also shows "Users" folder
ls -1
2021-04-26-14-45-37.zip	artisan			config			phpunit.xml		server.php
README.md		bootstrap		database		public			storage
Users			composer.json		db-dumps		resources		tests
app			composer.lock		package.json		routes			webpack.mix.js

# the "config" folder shows the expected files
$ ls config
app.php			broadcasting.php	database.php		logging.php		services.php
auth.php		cache.php		filesystems.php		mail.php		session.php
backup.php		cors.php		hashing.php		queue.php		view.php

# but under "Users" folder there also is a config folder
$ ls Users/path_to_source_code/
app		config		public		routes		tests
bootstrap	database	resources	storage

# but this folder is empty
$ ls Users/path_to_source_code/config
```

I extract using default unzip on MacOs (also tried p7zip).

### LINKED ISSUE #1299: "Links not supported" error when running "backup:monitor"
Running version 6.15.1 of laravel-backup on Laravel 8.36.2

Since the upgrade from 6.14.3 of laravel-backup the "backup:monitor" reports unhealthy backups due to the "local" filesystem not being reachable.
Backups are successfully written however.
There seems to be a problem with symlinks as there is an exception thrown "Links not supported" during execution.

I was able to circumvent this problem with the option `'links' => 'skip'` in my filesystems.php

```
'disks' => [

        'local' => [
            'driver' => 'local',
            'root' => storage_path('app'),
            'links' => 'skip'
        ],
```

That does not seem to be a good long term solution though.

### LINKED ISSUE #1312: --disable-notifications also disable encryption

### LINKED ISSUE #1357: File exclusion logic excludes files with same prefix

erikn69:
For V6
Closes #1357
Closes #1312
Closes #1298

```
PHPUnit 9.5.8 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.0.5
Configuration: /var/www/github/laravel-backup/phpunit.xml.dist

...............................................................  63 / 108 ( 58%)
.............................................                   108 / 108 (100%)

Time: 00:29.707, Memory: 38.00 MB

OK (108 tests, 2267 assertions)
```

@freekmurze  Maybe just merge this and close other PRs, so it can be merged to V7 for upgrade

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -6,7 +6,7 @@ jobs:
     tests:
         runs-on: ${{ matrix.os }}
         strategy:
-            fail-fast: true
+            fail-fast: false
             matrix:
                 os: [ubuntu-latest]
                 php: [8.0, 7.4, 7.3]
--- a/src/Tasks/Backup/BackupJob.php
+++ b/src/Tasks/Backup/BackupJob.php
@@ -214,7 +214,11 @@ class BackupJob
 
         consoleOutput()->info("Created zip containing {$zip->count()} files and directories. Size is {$zip->humanReadableSize()}");
 
-        $this->sendNotification(new BackupZipWasCreated($pathToZip));
+        if ($this->sendNotifications) {
+            $this->sendNotification(new BackupZipWasCreated($pathToZip));
+        } else {
+            app()->call('\Spatie\Backup\Listeners\EncryptBackupArchive@handle', ['event' => new BackupZipWasCreated($pathToZip)]);
+        }
 
         return $pathToZip;
     }
--- a/src/Tasks/Backup/FileSelection.php
+++ b/src/Tasks/Backup/FileSelection.php
@@ -131,8 +131,15 @@ class FileSelection
 
     protected function shouldExclude(string $path): bool
     {
+        $path = realpath($path);
+        if (is_dir($path)) {
+            $path .= '/';
+        }
         foreach ($this->excludeFilesAndDirectories as $excludedPath) {
-            if (Str::startsWith(realpath($path), $excludedPath)) {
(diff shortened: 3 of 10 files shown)


freekmurze:
Thanks!

PR #1365, opened 2021-08-10
Avoid create full path folders when config has relative_path

erikn69:
Try to fix 
Closes #1298

PR #1387, opened 2021-08-19
[v7] **Important** Avoid full path on db dumps

erikn69:
Avoid full path on db dumps

Revert bug added on #1365 (Sorry, my mistake)
Closes #1377


freekmurze:
Thanks!

PR #1411, opened 2021-10-03
Fix discord notification not getting defined username

fabegalo:


src/Notifications/Channels/Discord/DiscordMessage.php
```diff
@@ -113,7 +113,7 @@ public function fields(array $fields, bool $inline = true): self
     public function toArray(): array
     {
         return [
-            'username' => 'Laravel Backup',
+            'username' => $this->username,
```
freekmurze, line 116: Username isn't alway set, maybe change this to `$this->username ?? 'Laravel Backup'

PR #1414, opened 2021-10-04
[V6] Test for avoid full path on db file

erikn69:
Test for avoid questions like #1413, https://github.com/spatie/laravel-backup/discussions/1377#discussioncomment-1207602, and avoid break this again


freekmurze:
Thanks!

PR #1423, opened 2021-10-21
Fix discord notification message not using defined avatar url

pabueco:
Just a heads up, I didn't test if the avatar is actually missing. I just noticed that `$this->avatarUrl` isn't actually used in the `toArray()` method or anywhere else. So it seems like a custom avatar won't work.


freekmurze:
Thanks!

PR #1424, opened 2021-10-21
Replaced forDriver() if conditions with new php8.0 match expression

mo7zayed:
Since the package requires php8.0, I just added the new match for multiple conditions instead of repeating if conditions

Ref: https://www.php.net/manual/en/control-structures.match.php


WouterBrouwers:
@mo7zayed this PR has been selected as one of the winners of our Hacktoberfest contest, congratulations! 🎉
Can you share your Twitter handle with us so we can tweet about it?

mo7zayed:
@WouterBrouwers yah sure happy to share :), here it is `mo7azayed`

PR #1440, opened 2021-12-10
Fixed translated strings in Dutch

stephanbouman:
newest_backup_date and oldest_backup_date were not translated.


freekmurze:
Thanks!

PR #1460, opened 2022-01-21
Drop offered flysystem version that is uninstallable

abenerd:
Laravel 9 only supports league/flysystem v3, so offering v2 in the composer file is unnecessary as it can never be installed.


freekmurze:
Thanks!

Issue #1468, opened 2022-02-09
Backup status seems wrong when zip copy (to s3) fails 


smknstd:
**Versions**
PHP: 8.1.2
laravel: 8.78.1
laravel-backup: 7.3.1

When I setup a wrong bucket name, the zip file copy to s3 fails but it prints "Backup completed!" and return status is 0:

```
$ php artisan backup:run
Starting backup...
Dumping database ekip...
Determining files to backup...
Zipping 8 files and directories...
Created zip containing 8 files and directories. Size is 577.36 KB
Copying zip to disk named s3_backups...
Copying zip failed because: Error executing "PutObject" on "https://a-non-existing-bucket-name-in-the-given-region.s3.eu-central-1.amazonaws.com/Ekip/2022-02-09-09-40-25.zip"; AWS HTTP error: Client error: `PUT https://a-non-existing-bucket-name-in-the-given-region.s3.eu-central-1.amazonaws.com/Ekip/2022-02-09-09-40-25.zip` resulted in a `404 Not Found` response:
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NoSuchBucket</Code><Message>The specified bucket does not exist</Mes (truncated...)
 NoSuchBucket (client): The specified bucket does not exist - <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NoSuchBucket</Code><Message>The specified bucket does not exist</Message><BucketName>a-non-existing-bucket-name-in-the-given-region</BucketName><RequestId>W52FSE08ZNS*****</RequestId><HostId>st0yHWQe*****</HostId></Error>.
Backup completed!
```

```
$ echo $?
0
```

What do you think ?

PR #1469, opened 2022-02-09
Check if backup destination is reachable before trying to copy backup

smknstd:
close https://github.com/spatie/laravel-backup/issues/1468


smknstd:
not sure how to write a proper test for this one to make `isReachable()` returns false... any idea ?

freekmurze:
Thanks!

PR #1486, opened 2022-02-13
Fix issue where local backups were included in new backups

### LINKED ISSUE #1485: Previous local backups are included in backups since v8
I had an email this morning (thanks the scheduled health check command) to say that my backups were totalling more than 5gb. This was unusual since they backups are normally only around 15mb. I checked and the size of the backups were compounding starting at 300mb and doubling in size daily up until this morning being 2.6GB!

I updated to Laravel 9 and Backup 8 the other day so figured this is when the issue was introduced. After doing some digging, it's due to change made in Flysystem v3. The method in question is `BackupJob::directoriesUsedByBackupJob` where the wrong class name is used and upon fixing this there's a call to an undefined method `getAdapter`.

Screenshot 2022-02-13 at 14 33 05

Steps to reproduce:
1. Create a new laravel project `laravel new laravel-9`
2. Install spatie/laravel-backup `composer require spatie/laravel-backup`
3. Publish the config `php artisan vendor:publish --tag=backup-config`
4. Remove mysql from the databases key of the config
5. Remove mail from all notifications
6. Run the backup `php artisan backup:run`
7. Run this repeatedly to see the size of the backup double every time, show in the screenshot above

I've already tested a fix for this and will raise a PR.

Diagnost1X:
I have added a test which fails without the changes made in the subsequent commit.

Fixes #1485

### DIFF
--- a/src/Tasks/Backup/BackupJob.php
+++ b/src/Tasks/Backup/BackupJob.php
@@ -208,9 +208,9 @@ class BackupJob
     protected function directoriesUsedByBackupJob(): array
     {
         return $this->backupDestinations
-            ->filter(fn (BackupDestination $backupDestination) => $backupDestination->filesystemType() === 'local')
+            ->filter(fn (BackupDestination $backupDestination) => $backupDestination->filesystemType() === 'localfilesystemadapter')
             ->map(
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

PR #1487, opened 2022-02-18
Updates newest_backup_date and oldest_backup_date

opengis:
Updates newest_backup_date and oldest_backup_date keys in the french translation files to be Date instead of Size (Date instead of Taille).

Thanks for this great package (and all others you maintain)!


freekmurze:
Thanks!

PR #1493, opened 2022-02-23
Fix mistakes in cs/notifications.php

misakstvanu:
I changed some bad translations and typos.

Additionally - duplication on lines 40 and 43 caused emails not to have line with backup size (since the collection is filtered)


freekmurze:
Thanks!

PR #1494, opened 2022-02-28
Fix legacy backup config compatibility

bzarzuela:
Provides support for a problem I encountered in production where I used the schedule:list command on the console and wondered why it shows an unhealthy backup. I noticed that the disk configured was called local but mine was something else.

After investigating, I discovered that my config file used monitorBackups instead of monitor_backups. 

It seems that the config key was changed some time ago but I was unaware so I figure there might be others affected and thus, I created this pull request to help maintain backwards compatibility with those installations.

### DIFF
--- a/src/Commands/ListCommand.php
+++ b/src/Commands/ListCommand.php
@@ -19,7 +19,7 @@ class ListCommand extends BaseCommand
 
     public function handle()
     {
-        $statuses = BackupDestinationStatusFactory::createForMonitorConfig(config('backup.monitor_backups'));
+        $statuses = BackupDestinationStatusFactory::createForMonitorConfig(config('backup.monitor_backups', config('backup.monitorBackups')));
 
         $this->displayOverview($statuses)->displayFailures($statuses);
     }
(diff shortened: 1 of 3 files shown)


freekmurze:
Thank you for this, but I'm not going to merge it in, as I don't want to support the config formats of older package versions.

bzarzuela:
You're welcome! I didn't realize that change happened on v6. I must have started using this package on an even earlier version.

I guess another way of approaching this would be to show an error message if that key did not exist in the config file. That way, if there's someone else like me who missed updating the config key, they won't have to spend time tracking down the issue. I can create another PR with this approach if ever.

freekmurze:
I think I would like that approach better 👍

bzarzuela:
While working on this, I realized the problem cannot be handled by checking the existence of the config key because the Package Service Provider merges the default value from the config stub with the one published in the application. This is why I didn't get the type error in the createForMonitorConfig() function when I ran it in production. It still received an array, but the wrong one, in this case.

I could try and check for the existence of the old key and warn the user about it instead. I think it would still achieve my goal of saving someone else from figuring out why backup:run completes successfully but the output of backup:list shows an unhealthy output with the disk set to local.

PR #1495, opened 2022-03-01
Warns the user about old-style config files for backup:list and backup:monitor commands

bzarzuela:
Related to PR #1494 

Since that approach won't work as mentioned in my [comment ](https://github.com/spatie/laravel-backup/pull/1494#issuecomment-1055326350), this approach warns the user instead to ensure that his config key uses the snake_case format instead of camelCase.


AlexVanderbist:
Hi @bzarzuela, I'll merge this in as the final update for v8.x. Starting from 9.x I'll remove the checks again to keep the package clean and simple but I'll make sure to mention the config keys in the upgrading guide. Thanks for this!

PR #1499, opened 2022-03-02
Allow any characters in backup destination name

misakstvanu:
This allows navigating subfolders with "/"
...as discussed in https://github.com/spatie/laravel-backup/discussions/1497


freekmurze:
Thanks!

PR #1518, opened 2022-04-08
Use non-recursive files() method to test if destination is reachable

mike-healy:
The `allFiles()` method is recursive, and given that `isReachable` just tests if the destination is available and does not use the result of the method call, it seems unecessary to fetch recursively.

This PR swaps to the non-recursive `files()` method to do less work to determine the same answer.


freekmurze:
Nice, thank you!

PR #1533, opened 2022-06-03
Bug/1528 year week combo

### LINKED ISSUE #1528: Wrong Year-Week combination for first week of January
As you can see below the current `YW` format results in wrong year week combinations for end/start of year days/weeks. A curent example 2021/2022 - the first days of January 2022 would be added to the end of 2022 dates. so it would delete the wrong backups at the end of the year.

The solution would be to use the `formatLocalized()` method instead with the `%G%V` format. This would put the start of January days but in the end of 2021 week.
As the format logic is abstracted it would require changing all formats - but this should be easy as all others use straight year, month, day stuff.

```php
Carbon::parse('2021-12-31')->format('YW'); // 202152
Carbon::parse('2021-12-31')->formatLocalized('%G%V'); // 202152

Carbon::parse('2022-01-01')->format('YW'); // 202252
Carbon::parse('2022-01-01')->formatLocalized('%G%V'); // 202152

Carbon::parse('2022-12-31')->format('YW'); // 202252
Carbon::parse('2022-12-31')->formatLocalized('%G%V'); // 202252
```

For all groups it would be:
```php
formatLocalized('%Y%m%d');
formatLocalized('%G%V');
formatLocalized('%Y%m');
formatLocalized('%Y');
```

And as we all like a good StackOverflow link, here you go: https://stackoverflow.com/a/64641985/4907524 😉

Nielsvanpach:
I only found `YW` be used once in the codebase, when deleting old backups.
Also,`formatLocalized()` is deprecated, `isoFormat()`is suggested as alternative.

Following the examples in #1528, `GW` gave the same result as the `%G%V` suggestion.
```php
Carbon::parse('2022-01-01')->formatLocalized('%G%V'); // 202152
Carbon::parse('2022-01-01')->isoFormat('GW'); // 202152
```

Closes #1528

### DIFF
--- a/src/Tasks/Cleanup/Strategies/DefaultStrategy.php
+++ b/src/Tasks/Cleanup/Strategies/DefaultStrategy.php
@@ -26,7 +26,7 @@ class DefaultStrategy extends CleanupStrategy
         });
 
         $backupsPerPeriod['daily'] = $this->groupByDateFormat($backupsPerPeriod['daily'], 'Ymd');
-        $backupsPerPeriod['weekly'] = $this->groupByDateIsoFormat($backupsPerPeriod['weekly'], 'YW');
+        $backupsPerPeriod['weekly'] = $this->groupByDateIsoFormat($backupsPerPeriod['weekly'], 'GW');
         $backupsPerPeriod['monthly'] = $this->groupByDateFormat($backupsPerPeriod['monthly'], 'Ym');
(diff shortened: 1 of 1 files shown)


Gummibeer:
I think that a few cases should be added here. As the new grouping is theoretically correct but could be problematic with the current clean up procedure when I got the behavior right.

Gummibeer:
This isn't fixed/merged - so should still be open.

PR #1538, opened 2022-06-17
Add a note about scenarios where backups could be lost unexpectedly

wmather:
I have been using this package successfully for over 7 years, but i've had an issue for the last year that has happened in two different codebases. Older backups were being deleted that didn't respect the config choices.

I (incorrectly) always assumed the backup date was determined from the file name, but this is not the case and explains why the backups were being deleted in these scenarios:

1. I moved the backups from DO to GCP and naturally all the timestamps were refreshed when I uploaded them. After `keep_all_backups_for_days`, all older backups would be wiped except one.
2. On another project (with versioning applied in the GCP bucket), after a while it became apparent that our config choices were not suitable and I wanted to increase the amount of monthly and yearly backups. So I started restoring the older deleted backups, which gave them a last modified date of now. Similarly, after `keep_all_backups_for_days` all but one would be deleted.

I believe that i'll need to write a custom strategy because there is no way I can change the last modified dates of these files to something in the past .

This PR puts a warning in the documentation to highlight this aspect, because I have been chasing this for a long time and have only just figured it out! (and if I didn't have versioning enabled, they'd be gone forever)

### DIFF
--- a/docs/cleaning-up-old-backups/overview.md
+++ b/docs/cleaning-up-old-backups/overview.md
@@ -80,6 +80,12 @@ older than those covered by rule #2
 
 Of course the numbers used in the default configuration can be adjusted to suit your own needs.
 
+> It is worth noting that the date of a backup is determined by the last modified date of the file. This can potentially mean unexpected backup deletions occur in the following scenarios:
+> - You move your backups to a new storage provider
+> - You restore a previously deleted backup, if the storage provider has file versioning capability
+> 
+> The solution would be to create your own strategy that does not use the last modified date of the file, but instead extracts the date from the file name
+
 ## Creating your own strategy
 
 If you're requirements are not covered by the `DefaultStrategy`, you can create your own custom strategy. 


wmather:
Let me qualify this a bit more, the package does try to use the file name date: https://github.com/spatie/laravel-backup/blob/1364a6aed65aef8a347b38bbc83ac74c19897246/src/BackupDestination/Backup.php#L44-L50

Except where you have set a `destination.file_prefix` in the config (or naturally change the file name structure), because the file name date format is statically defined and there is no control to change this https://github.com/spatie/laravel-backup/blob/1364a6aed65aef8a347b38bbc83ac74c19897246/src/Tasks/Backup/BackupJob.php#L26

bbashy:
Good to note I think.

Maybe also add something about if you move or restore backups, you should make sure the last modified date is within the config ranges.

rsync can preserve the dates in transfers or the Linux `touch` command can be used to set the date, if required.

docs/cleaning-up-old-backups/overview.md
```diff
@@ -80,6 +80,12 @@ older than those covered by rule #2
 
 Of course the numbers used in the default configuration can be adjusted to suit your own needs.
 
+> It is worth noting that the date of a backup is determined by the last modified date of the file if you have added value in the `destination.filename_prefix` config or have changed the file name from the default structure. This can potentially mean unexpected backup deletions occur in the following scenarios:
```
Jamesking56, line 83:
```suggestion
> It is worth noting that the date of a backup is determined by the last modified date of the file. If you have added a value in the `destination.filename_prefix` config or have changed the file name from the default structure, this can potentially mean unexpected backup deletions occur in the following scenarios:
```

docs/cleaning-up-old-backups/overview.md
```diff
@@ -80,6 +80,12 @@ older than those covered by rule #2
...
+> - You move your backups to a new storage provider
+> - You restore a previously deleted backup, if the storage provider has file versioning capability
+> 
+> The solution would be to create your own strategy that does not use the last modified date of the file, but instead extracts the date from the file name, or to not use a file prefix and potentially use multiple disks with different folders
```
Jamesking56, line 87:
```suggestion
> The solution would be to create your own strategy that does not use the last modified date of the file, but instead extracts the date from the file name, or to not use a file prefix and potentially use multiple disks with different folder structures.
```

PR #1540, opened 2022-06-30
docs(setup): update default config

tricki:
Used current source of https://github.com/spatie/laravel-backup/blob/main/config/backup.php


freekmurze:
Thanks!

PR #1544, opened 2022-07-05
Use empty string for Discord defaults

bbashy:
Using `null` as the defaults didn't work when just entering the webhook URL. Most of the time, you will make a webhook for each notification and set name/avatar there.

This is with it being `null`:
```
Spatie\Backup\Notifications\Channels\Discord\DiscordMessage::from(): Argument #1 ($username) must be of type string, null given
```

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -191,9 +191,15 @@ return [
         'discord' => [
             'webhook_url' => '',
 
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #1552, opened 2022-07-27
changed symfony console & finder versions

mudassar1:
updated symfony console to 6.1.2 & 6.1.0


freekmurze:
Thanks, but passing as there is no real added value here.

PR #1560, opened 2022-08-04
Improve output messages for backup:monitor command

intrepidws:
The output messages for this command are missing the name of the backup. This PR proposes to fix that. So instead of getting output like this:

```
The backups on s3 are considered healthy.
The backups on s3 are considered healthy.
The backups on r2 are considered healthy.
The backups on r2 are considered healthy.
```

You would instead get this:

```
The Project One backups on the s3 disk are considered healthy.
The Project Two backups on the s3 disk are considered healthy.
The Project Three backups on the r2 disk are considered healthy.
The Project Four backups on the r2 disk are considered healthy.
```


freekmurze:
Thanks!

PR #1571, opened 2022-09-11
Update adding-extra-notification-channels.md

llabbasmkhll:
The extra notification channel was unclear to me and caused me some reverse engineering to figure it out.
I think this will make it more meaningful.


freekmurze:
Thank you!

PR #1575, opened 2022-09-23
Do not delete any oldest backup using null value

kmpratik:


freekmurze:
It will never delete the oldest backup with the current code.

PR #1624, opened 2022-12-23
Allow to backup databases that are not in backup config

applyACS:


freekmurze:
Could you explain why this change is needed? 

If it's because of a bug, could you add a tests that would fail without this change?

applyACS:
You currently can not backup manually a DB that it's connection is not set in config backup.source.databases. I opened a discussion some time ago: https://github.com/spatie/laravel-backup/discussions/1593
So I want to do a backup of a database that I don't want to backup regularly. I don't want to include this in my daily backup.
This fixes the issue.

freekmurze:
Could you add a test to make sure this change works correctly?

PR #1634, opened 2023-01-25
fix default Discord username

bbashy:
My Discord notifications stopped working after 9th Jan 2023.

It appears it's because of the default username not being set properly. This change fixes it (username is required, default is Laravel Backup).

Discord returns this error now;
```
"{"username": ["Must be between 1 and 80 in length.", "Username cannot be \"\""]}"
// vendor/spatie/laravel-backup/src/Notifications/Channels/Discord/DiscordChannel.php:16
```

`username` can be set to null or '' so it won't break any existing configs.

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -192,14 +192,15 @@ return [
             'webhook_url' => '',
 
             /*
-             * If this is an empty string, the name field on the webhook will be used.
+             * If this is null or an empty string, "Laravel Backup" will be used.
+             * Must be between 1 and 80 in length.
              */
-            'username' => '',
+            'username' => null,
 
             /*
-             * If this is an empty string, the avatar on the webhook will be used.
+             * If this is null or an empty string, the avatar on the webhook will be used.
              */
-            'avatar_url' => '',
+            'avatar_url' => null,
         ],
     ],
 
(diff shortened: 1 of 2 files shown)


config/backup.php
```diff
@@ -192,14 +192,15 @@
             'webhook_url' => '',
 
             /*
-             * If this is an empty string, the name field on the webhook will be used.
+             * If this is null or an empty string, "Laravel Backup" will be used.
```
freekmurze, line 195: Let's revert these changes

src/Notifications/Channels/Discord/DiscordMessage.php
```diff
@@ -28,9 +28,11 @@ class DiscordMessage
 
     protected string $url = '';
 
-    public function from(string $username, string $avatarUrl = null): self
+    public function from(string $username = null, string $avatarUrl = null): self
     {
-        $this->username = $username;
+        if (! empty(trim($username))) {
```
freekmurze, line 33: Let's revert all of the changes in this file. 

Add a check if the username is equal to an empty string, and if so, set the username to Laravel Backup

freekmurze (changes requested):
Let's implement this with less changes.

bbashy:
I'll leave this then. If the username is null or '', it currently doesn't work.

I've set the username in the config which fixed it.

PR #1640, opened 2023-02-03
Update DbDumperFactory.php

nikosv:
I am adding $dbConfig['connect_via_database'] in setDbName() as the package doesn't work with pgbouncer configuration due to the changes described here: https://github.com/laravel/framework/pull/43542

Summary: Since Laravel 9 in case you are using pgbouncer for pooling you need to add the name of the pool on a separate config variable (connect_via_database) instead of database that was used in earlier versions.


freekmurze:
Thank you!

PR #1648, opened 2023-03-18
avoid block other listeners from fired

bangnokia:
Hi! This pull request moves the rescue for sending notifications from the EventHandler to avoid blocking other listeners from being fired when listening to a specific event from this package.

For instance, when writing a plugin for this package, we may listen to a particular event being fired to handle extra work, like BackupWasSuccessful. However, if the notification sending fails (for example, due to a mail error or a changed Slack API key), it may block other listeners from being fired when using the queue sync.

### DIFF
--- a/src/Notifications/EventHandler.php
+++ b/src/Notifications/EventHandler.php
@@ -27,7 +27,10 @@ class EventHandler
 
             $notification = $this->determineNotification($event);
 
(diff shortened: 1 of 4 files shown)


freekmurze:
Thanks! I'm going to leave things around this as they are for now.

PR #1654, opened 2023-04-01
Update EventHandler.php

1cbyc:
You guys wrote a really good code here. For this to stay up for 3 years. However, there are a few things I think can be adjusted. 

If the determineNotification() method is unable to find a matching notification class, it throws a ‘NotificationCouldNotBeSent’ exception. I think this exception is not caught or handled anywhere, so it would be better to add a try-catch block to handle this exception. Also, it’s clear the determineNotification() method should have a return type declaration to ensure that it returns an instance of the Notification class. You might want to rename some method names like determineNotifiable() and determineNotification() as they do not seem very clear.

### DIFF
--- a/src/Notifications/EventHandler.php
+++ b/src/Notifications/EventHandler.php
@@ -22,23 +22,27 @@ class EventHandler
 
     public function subscribe(Dispatcher $events): void
     {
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks! I'm going to keep the current code as we currently have no issues with it.

PR #1655, opened 2023-04-01
Update HealthCheck.php

1cbyc:
Yo! I ONLY just tried to make the code more readable. I only made four changes that might be helpful. Seeing that the `checkHealth` method's return type is not defined. Noticed that the `fail` method should create a new `InvalidHealthCheck` exception object and throw it, instead of calling the `because` method. Expected it to be `void`. Also, `name` method should be renamed to `getName`, according to PSR-1 naming conventions. However, the `failUnless` method should negate the `$condition` parameter in the `if` statement to match its name.


freekmurze:
Thanks! I'm going to leave things as they are now, as some of these changes will break existing installations.

PR #1661, opened 2023-04-25
Show accurate list of supported drivers when throwing `CannotCreateDbDumper` exception

dejagersh:
When specifying a non-existent connection in `config('backup.backup.source.databases')`, it will throw the `CannotCreateDbDumper` exception. The current exception message states that only the default Laravel connection names are supported, which is misleading. In reality, any connection defined in `config('database.connections')` can be used, regardless of whether it is a standard Laravel connection name or not.

This PR modifies the exception message to display the actual available connection names that can be used. This will help users better understand which connections are valid options and reduce confusion.

image


src/Exceptions/InvalidConfiguration.php
```diff
@@ -1,13 +0,0 @@
-<?php
-
-namespace Spatie\Backup\Exceptions;
-
-use Exception;
-
-class InvalidConfiguration extends Exception
```
dejagersh, line 7: I tried to find where this was used. I can't find it. It seems unused to me. Or is it somehow used somewhere?

freekmurze: I'll remove this one later.

freekmurze:
Thanks!

PR #1683, opened 2023-07-10
fix: relative_path too short etc: /app

jacky158:
fix error when run laravel-backup in docker container 

etc: relative_path=> '/app'

relative_path trims `storage/app/backup` to `storagebackup`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "spatie/laravel-backup",
+    "name": "metafox/laravel-backup",
     "description": "A Laravel package to backup your application",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "spatie/laravel-backup",
+    "name": "metafox/laravel-backup",
```
erikn69, line 2: ??

freekmurze:
Could you add a test for this and correct vendor in package.json?

freekmurze:
Closing due to inactivity.

Issue #752, opened 2018-09-16
Feature request for "retry"


1e4:
I was recently playing around with the Backup tool and it works great, the one problem I have though is if the upload fails for whatever reason I have to rerun the backup, which then zips up all the files again.

It would be great if there was a retry command which simply retries any failed backups.

Thanks for the package 👍

freekmurze:
Because the currenr setup works for the vast majority of users and I like to keep the code base simple, I’m probably not going to include such a feature soon.

PR #1684, opened 2023-07-24
Add a retry feature to the `backup` and `clean` commands

prestamodule:
Hello,

Due to random errors with an S3-compatible backend, I had to create a little retry mechanism for both of these operations to avoid getting spammed of failed backups/cleanups notifications when a simple retry a few seconds later would do the trick.

I believe this is what was requested in #752 as well.

Let me know if anything is missing or incomplete :-)


src/Commands/BackupCommand.php
```diff
@@ -47,6 +51,12 @@ public function handle()
                 $backupJob->setFilename($this->option('filename'));
             }
 
+            if ($this->option('tries')) {
+                $this->tries = (int)$this->option('tries');
+            } elseif (!empty(config('backup.backup.tries'))) {
```
freekmurze, line 56: I really dislike `else` statements. Could you refactor this to a dedicated method with early returns?

src/Commands/BackupCommand.php
```diff
@@ -59,6 +69,15 @@ public function handle()
 
             consoleOutput()->comment('Backup completed!');
         } catch (Exception $exception) {
+            if ($this->tries > 1 && $this->currentTry < $this->tries) {
```
freekmurze, line 72: Conditionals like this are not that readable. Could you refactor this to a dedicated method with a good name, that returns a boolean and uses early returns.

src/Commands/CleanupCommand.php
```diff
@@ -42,6 +52,15 @@ public function handle()
 
             consoleOutput()->comment('Cleanup completed!');
         } catch (Exception $exception) {
+            if ($this->tries > 1 && $this->currentTry < $this->tries) {
```
freekmurze, line 55: Refactor to a the conditional to a dedicated method.

src/Commands/BackupCommand.php
```diff
@@ -59,6 +69,15 @@ public function handle()
...
+                if (!empty(config('backup.backup.retry_delay'))) {
+                    sleep((int)config('backup.backup.retry_delay'));
```
freekmurze, line 74: Instead of using `sleep` use the new sleep helper that is easier to test: https://www.amitmerchant.com/the-new-sleep-helper-in-laravel-10/

freekmurze (changes requested):

prestamodule:
Hi,

I just made some changes per your comments, let me know if you want anything else updated :-)

src/Helpers/functions.php
```diff
@@ -6,3 +6,8 @@ function consoleOutput(): ConsoleOutput
 {
     return app(ConsoleOutput::class);
 }
+
+function isSleepHelperAvailable()
```
freekmurze, line 10: Let's just update the minimum required version for the package, so we can dispatch this function and all it's usages.

freekmurze (changes requested):

src/Helpers/functions.php
prestamodule: Alright, I've just removed it

freekmurze:
The tests are failing for this one, could you take a look

prestamodule:
I believe this is due to the fact that the `Sleep` helper was only introduced in Laravel 10.10, and not with the 10.0 release ?
https://github.com/laravel/framework/releases/tag/v10.10.0

prestamodule:
Hi,

Just bumped the requirements to >= 10.10.0, let me know if this is good now :-)

prestamodule:
Hey @freekmurze, is there anything else I need to add or change to this PR?
It seems the configured workflow requires your approval to run the tests following the dependency update.

freekmurze:
Thanks!

prestamodule:
Thank you :)

PR #1694, opened 2023-08-09
Fix: relative_path too short, example: /app

erikn69:
https://github.com/spatie/laravel-backup/pull/1683#issuecomment-1671017968

This problem should be solved


freekmurze:
Thanks!

PR #1697, opened 2023-08-13
Update Laravel & Backup Server docs to latest version

rico-vz:
I've updated the Laravel docs and the Laravel Backup Server docs to their latest/main available version.


freekmurze:
Thank you!

PR #1706, opened 2023-08-25
Return `Command` constants instead of integers in commands

joshbonnick:
## Summary
This PR converts returning integer exit codes to returning the constants defined in [Symfony Commands class](https://github.com/symfony/console/blob/6.3/Command/Command.php?plain=1#L38), making the code more expressive and easier to understand.

## Example
```php
class MyCommand extends Command
{
    public function handle(): int 
    {
        // do something that fails...

        return 1;
    }
}
```

```php
class MyCommand extends Command 
{
    public function handle(): int 
    {
        // do something that fails...

        return static::FAILURE;
    }
}
```

## Additional Changes

- Return success exit code where command is returning failure or void, ensuring the `handle` methods always returns an integer, never void.
- Declare return type `int` on `handle` methods.

## Possible Breaking Changes
If the user is extending one of the commands the package provides and are overriding the `handle` method without the `int` return type an exception will be thrown. I can remove the return type declaration added if this is a concern.

### DIFF
--- a/src/Commands/BackupCommand.php
+++ b/src/Commands/BackupCommand.php
@@ -16,7 +16,7 @@ class BackupCommand extends BaseCommand
 
     protected $description = 'Run the backup.';
 
-    public function handle()
+    public function handle(): int
     {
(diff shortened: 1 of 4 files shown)


freekmurze:
Thanks!

PR #1707, opened 2023-08-25
Send one notification on backup failure

### LINKED ISSUE #1662: I receive two email notifications whenever a backup fails
I have an odd, longstanding problem that I simply cannot figure out. Whenever a backup fails, it will always send two identical email notifications at the exact same time:

CleanShot 2023-04-26 at 18 29 15@2x

It doesn't matter if the error is due to a timeout, or a S3 issue, I will always get two emails in the event of a failure.

My configuration is set up with a single email:

CleanShot 2023-04-26 at 18 34 32@2x

And the backup is configured to run once, but it does have 'runInBackground()' specified:

CleanShot 2023-04-26 at 18 35 43@2x

I'm on version 8.1.7, and PHP 8.2 and MySQL 8.0.32.

The _only_ thing I can think of is that somehow the `backup:run` job is running twice and is actually the _cause_ of the failures, explaining multiple emails. But not sure how that could be happening.

joshbonnick:
Fixes #1662

## Summary
This PR addresses an issue with exception handling in [BackupJob](https://github.com/spatie/laravel-backup/blob/f4a2abc673ed61447ca748d02e9cc8be16e1c485/src/Tasks/Backup/BackupJob.php#L172C13-L172C13). 

As both the `BackupCommand` command and `BackupJob` are handling exceptions the result is both send the failure notification when they deem the process to have failed.

With these changes `BackupCommand` will exclusively handle notifying the user if a failure scenario occurs. Due to the command being able to retry itself it makes the most sense to allow **it** to determine when it considers the process to have failed. 

--------

Before these changes the `copyToBackupDestinations` method of `BackupJob` would fire the backup failed event and pass the destination as a parameter. 

https://github.com/spatie/laravel-backup/blob/f4a2abc673ed61447ca748d02e9cc8be16e1c485/src/Tasks/Backup/BackupJob.php#L300

To ensure behavior remains the same, I have added the `BackupFailed` exception which acts as a decorator with a public `$backupDestination` property which is used to set the same property in the backup failed event.

The backup destination property is set by chaining on the exception.

https://github.com/spatie/laravel-backup/blob/c394924f2016f709913e7b9894a7ed17898bdd4a/src/Tasks/Backup/BackupJob.php#L304

The notification receives the original exception for the user to debug.

https://github.com/spatie/laravel-backup/blob/dc70da6abab42f003eaa873bba792f82ffbda92a/src/Commands/BackupCommand.php#L84

## Possible Concerns

- `BackupJob` still handles success notifications and informative (including error) output to the console, we would need to completely decouple the job from the command to change this behavior, which I do not believe has any benefit to the package.
- PHPStan raised an issue with `getPrevious()` returning `Throwable|null` as `BackupHasFailed` event accepts `Exception` - to fix this an `@method` has been added to `BackupFailed`. This can be replaced with a `previous(): Exception` method if you prefer.

### DIFF
--- a/src/Commands/BackupCommand.php
+++ b/src/Commands/BackupCommand.php
@@ -74,6 +74,8 @@ class BackupCommand extends BaseCommand
                 return $this->handle();
             }
 
+            consoleOutput()->error("Backup failed because: {$exception->getMessage()}.");
+
             report($exception);
 
             if (! $disableNotifications) {
--- a/src/Commands/BackupCommand.php
+++ b/src/Commands/BackupCommand.php
@@ -4,6 +4,7 @@ namespace Spatie\Backup\Commands;
 
 use Exception;
 use Spatie\Backup\Events\BackupHasFailed;
+use Spatie\Backup\Exceptions\BackupFailed;
 use Spatie\Backup\Exceptions\InvalidCommand;
 use Spatie\Backup\Tasks\Backup\BackupJobFactory;
 use Spatie\Backup\Traits\Retryable;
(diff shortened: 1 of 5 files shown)


freekmurze:
Thanks!

PR #1712, opened 2023-09-15
Move registering events to "booted" rather than "registered"

modstore:
As it is, events aren't properly registered. If you run `artisan event:list` the events won't show. But moving to booted instead, will register the events correctly.


freekmurze:
Thanks!

PR #1714, opened 2023-09-18
Update `databases` key to use `DB_CONNECTION` from .env

maskeynihal:
Utilizing `env('DB_CONNECTION')` enables the package to function seamlessly without requiring any adjustments to the `backup.php` configuration file. It automatically derives the database connection information from the corresponding environment variable, ensuring flexibility in database selection based on the environment.


freekmurze:
I think this is a good idea, but it could be considered a breaking change, as there are situations this change will result in a different db being backed up. I'll revisit this when creating a new major version of this package.

PR #1717, opened 2023-10-02
Possibility of adding timestamp to database dump filename

boryn:
At this moment, after unzipping several dumps, the db filename always is the same - which makes it difficult to do any comparisons between them, and we are unaware when they have been taken.

With this PR, you can choose a timestamp format (through the `'database_dump_file_timestamp_format'` config option), which will be inserted into the db dump filename and now it's clear when it was taken, eg.:
`mysql-PROJECT-2023-10-02-07-00-00.sql`

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -98,6 +98,11 @@ return [
         'database_dump_compressor' => null,
 
         /*
+         * If specified, the database dumped file name will contain a timestamp (e.g.: 'Y-m-d-H-i-s').
+         */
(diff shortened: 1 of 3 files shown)


freekmurze:
Thanks! Could you add a test for this?

boryn:
I'll try :) Do you mean creating a new test with this option set?

boryn:
Ok, @freekmurze I have just added the `'appends timestamp to database backup file name'` test for this.

freekmurze:
Thanks!

PR #1718, opened 2023-10-02
Possibility of choosing zip compression method

boryn:
This PR adds possibility of choosing zip compression method and its level through these config options:
- `['destination']['compression_method']`
- `['destination']['compression_level']`

Now it's possible to select the specific compression method like `ZipArchive::CM_DEFLATE` or `ZipArchive::CM_XZ` or not use compression at all with `ZipArchive::CM_STORE`. 

It's possible to experiment and choose optimal compression for one's needs, which impacts memory/cpu usage plus time consumption and finally file size. 

In case of using only db dumps, I have decided to directly compress .sql with `GzipCompressor::class` and not compress .zip archive, which gives me less CPU usage and still acceptable file size.

### DIFF
--- a/src/Tasks/Backup/Zip.php
+++ b/src/Tasks/Backup/Zip.php
@@ -106,9 +106,9 @@ class Zip
             }
 
             if (is_file($file)) {
-                $this->zipFile->addFile($file, ltrim($nameInZip ?: '', DIRECTORY_SEPARATOR));
+                $this->zipFile->addFile($file, ltrim($nameInZip, DIRECTORY_SEPARATOR)).PHP_EOL;
 
-                if ($nameInZip && is_int($compressionMethod)) {
+                if (is_int($compressionMethod)) {
                     $this->zipFile->setCompressionName(
                         ltrim($nameInZip, DIRECTORY_SEPARATOR),
                         $compressionMethod,
--- a/src/Tasks/Backup/Zip.php
+++ b/src/Tasks/Backup/Zip.php
@@ -97,6 +97,9 @@ class Zip
             $files = [$files];
         }
 
+        $compressionMethod = config('backup.backup.destination.compression_method', null);
+        $compressionLevel = config('backup.backup.destination.compression_level', 9);
+
         foreach ($files as $file) {
             if (is_dir($file)) {
                 $this->zipFile->addEmptyDir(ltrim($nameInZip ?: $file, DIRECTORY_SEPARATOR));
@@ -104,6 +107,14 @@ class Zip
 
             if (is_file($file)) {
                 $this->zipFile->addFile($file, ltrim($nameInZip, DIRECTORY_SEPARATOR)).PHP_EOL;
+
+                if (is_int($compressionMethod)) {
+                    $this->zipFile->setCompressionName(
+                        ltrim($nameInZip, DIRECTORY_SEPARATOR),
+                        $compressionMethod,
+                        $compressionLevel
+                    );
+                }
             }
             $this->fileCount++;
         }
--- a/src/Tasks/Backup/Zip.php
+++ b/src/Tasks/Backup/Zip.php
@@ -110,7 +110,7 @@ class Zip
 
                 if (is_int($compressionMethod)) {
                     $this->zipFile->setCompressionName(
-                        ltrim($nameInZip, DIRECTORY_SEPARATOR),
+                        ltrim($nameInZip ?: $file, DIRECTORY_SEPARATOR),
                         $compressionMethod,
                         $compressionLevel
                     );
--- a/src/Tasks/Backup/Zip.php
+++ b/src/Tasks/Backup/Zip.php
@@ -106,7 +106,7 @@ class Zip
             }
 
             if (is_file($file)) {
-                $this->zipFile->addFile($file, ltrim($nameInZip, DIRECTORY_SEPARATOR)).PHP_EOL;
+                $this->zipFile->addFile($file, ltrim($nameInZip, DIRECTORY_SEPARATOR));
 
                 if (is_int($compressionMethod)) {
                     $this->zipFile->setCompressionName(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All notable changes to `laravel-backup` will be documented in this file.
 
+## 8.3.3 - 2023-09-04
+
+### What's Changed
+
+- Added missing PT and PT/BR translations by @helderneves91 in https://github.com/spatie/laravel-backup/pull/1709
+
+**Full Changelog**: https://github.com/spatie/laravel-backup/compare/8.3.2...8.3.3
+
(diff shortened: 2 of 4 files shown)


boryn:
@freekmurze, added `'can use different compression methods for backup file'` test

freekmurze:
Seems like the tests are failing, could you look at that?

erikn69:
- @boryn hi, i fix the tests: https://github.com/boryn/laravel-backup/pull/1 [/github/tests/results](https://github.com/erikn69/laravel-backup/actions/runs/6472352704/job/17572816514)

boryn:
@erikn69 Thank you!

PS. Why $nameInZip may be not present at all?

erikn69:
`$filepath` is used if `$entryname` is empty 
https://www.php.net/manual/es/ziparchive.addfile.php
```php
public ZipArchive::addFile(
    string $filepath,
    string $entryname = "",
    int $start = 0,
    int $length = 0,
    int $flags = ZipArchive::FL_OVERWRITE
): bool
```

src/Tasks/Backup/Zip.php
```diff
@@ -97,13 +97,24 @@ public function add(string | iterable $files, string $nameInZip = null): self
...
                 $this->zipFile->addEmptyDir(ltrim($nameInZip ?: $file, DIRECTORY_SEPARATOR));
             }
 
             if (is_file($file)) {
-                $this->zipFile->addFile($file, ltrim($nameInZip, DIRECTORY_SEPARATOR)).PHP_EOL;
+                $this->zipFile->addFile($file, ltrim($nameInZip ?: '', DIRECTORY_SEPARATOR));
+
+                if ($nameInZip && is_int($compressionMethod)) {
+                    $this->zipFile->setCompressionName(
+                        ltrim($nameInZip, DIRECTORY_SEPARATOR),
```
erikn69, line 113: now that I think about it better, maybe
`ltrim($nameInZip ?: '', DIRECTORY_SEPARATOR) ?: $file,`

boryn: I think so, as we'd rather use compression as well for the case where `$filepath` is used. I'd need to make some more investigation here.

src/Tasks/Backup/Zip.php
```diff
@@ -97,13 +97,24 @@ public function add(string | iterable $files, string $nameInZip = null): self
...
+        $compressionMethod = config('backup.backup.destination.compression_method', null);
+        $compressionLevel = config('backup.backup.destination.compression_level', 9);
+
         foreach ($files as $file) {
             if (is_dir($file)) {
                 $this->zipFile->addEmptyDir(ltrim($nameInZip ?: $file, DIRECTORY_SEPARATOR));
             }
 
             if (is_file($file)) {
                 $this->zipFile->addFile($file, ltrim($nameInZip, DIRECTORY_SEPARATOR)).PHP_EOL;
```
erikn69, line 109: Hi, what `.PHP_EOL`  do?? it is ouside

boryn:
Ok, @freekmurze I have fixed this issue. 

It was connected to not using `$nameInZip`. Now the compression works properly with or without `$nameInZip` parameter.

src/Tasks/Backup/Zip.php
boryn: I was wondering as well what it's doing there, but it is in the main branch:
https://github.com/spatie/laravel-backup/blob/f81e8495a7944eb9684580d99f40bbdb4b3f7f93/src/Tasks/Backup/Zip.php#L106C88-L106C88

and I have decided to leave it untouched

erikn69: Most likely it was a typo.

boryn: Ok, removed unnecessary `.PHP_EOL`. It's been sitting there since 2016 :)

freekmurze:
Thanks!

siarheipashkevich:
Hi @boryn thanks for your PR's. Could you please suggest which type of compression is better to use only for DB dump.

Server with database:
MySQL 8.0.34
image

Server with app where backup is doing:
php 8.2.11
image


Right now I have scheduler:
`$schedule->command('backup:run --only-db')->dailyAt('04:30');`

and it takes almost 1 hour of time

Screenshot from 2023-10-17 21-01-31

the size of the archive is is `3.67GB` and `mysql-forge.sql` is ~`30GB`.

The config `backup.php` is default but I added one option to the `database.php` config is:
image

Maybe I can decrease the time of the dump process and decrease the size?

boryn:
Hi @siarheipashkevich!

For the only db dump, I have gone with gzip compression directly at db dump, so at `backup.php` I have:
`'database_dump_compressor' => Spatie\DbDumper\Compressors\GzipCompressor::class,`
(the default compression level for gzip is 6 and for me it is a good compromise between speed and compression ratio)

and for ZIP I don't use any compression, so I have then: 
`'compression_method' => ZipArchive::CM_STORE,`

This configuration gives me acceptable compression ratio with not CPU hogging. 

Actually, it depends what you need. For better ZIP compression (decreased file size, but definitely more compression time, memory and CPU) you may try `ZipArchive::CM_XZ` or `ZipArchive::CM_BZIP2`, or as well experiment with `CM_LZMA` or `CM_LZ77`.

boryn:
Yes, though `compression_level` for zip is not relevant when using `ZipArchive::CM_STORE` for just storing the zip file without compression. 

If you want to play with compression level for gzip, you'd need to extend and create your own copy of `Spatie\DbDumper\Compressors\GzipCompressor::class`, where you would specify:

```php
public function useCommand(): string
{
    return 'gzip -3';
}
```

and you can play with different compression levels by changing the gzip parameter like -1, -2, -3, -6, -9, etc.

My tests show, that using gzip with the default compression (which is 6) does not hog the CPU and users can easily use the system. And using gzip should be quicker than the default zipping options with compression at level 9.

siarheipashkevich:
@boryn thanks, I trust your configuration :smile: 

Sorry for the stupid question, is it ok and it is about your recommendation? :
image

boryn:
Yes, but please run the backup on the development / local machine first :) 

Let us know how long it took and how the cpu was used.

PS. You may even remove the `compression_level` key

siarheipashkevich:
@boryn 

app server:
image

database server:
image

As I can see the CPU is higher in app server right now.

boryn:
Yes, interesting. Maybe try to experiment with extending the gzip class and using compression like -3 ?

siarheipashkevich:
Hi @boryn, there is screens from my production:

database server:
image

app server:
image

I see some spike of  the CPU usage in app server but I think it's not critical.
And also time decreased by 2 times. Thanks for your recommendations :+1: 

Do you have an experience of tuning MySQL or creating replication for production database?

boryn:
Hi @siarheipashkevich 

It's so great that the time has reduced 2x! Unfortunately I have actually no experience with replication of MySQL.

PR #1726, opened 2023-10-26
Use defined config

xolf:
I added `spatie/laravel-backup` to my project and was curious why it wasn't using my default database and filesystem for the backup.

With replacing the hardcoded settings for the database and filesystem, the packe might be instantly used out of the box.


freekmurze:
I think it's a good change, but right now this might change behaviour for some users.

I'll take a look again when creating the next major version.

PR #1728, opened 2023-10-31
Improve strategy comments

iwasherefirst2:
When reading the config the first time, I was very confused about the term "daily/weekly/monthly/yearly backup". 

The confusion arises because the terms imply a specific backup frequency but in reality, the package uses these terms to describe how long backups should be retained in each category, regardless of the actual frequency at which backups are created.

For better clarity, I would recommend to update the comments as proposed.

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -286,7 +286,7 @@ return [
         'default_strategy' => [
 
             /*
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

Issue #613, opened 2018-01-25
How can I cancel some notifications?


itzikbenh:
Hey, when I do this I get an error and backup is failed. 
```
'notifications' => [
    \Spatie\Backup\Notifications\Notifications\BackupHasFailed::class         => ['mail'],
    \Spatie\Backup\Notifications\Notifications\UnhealthyBackupWasFound::class => ['mail'],
    \Spatie\Backup\Notifications\Notifications\CleanupHasFailed::class        => ['mail'],
    //\Spatie\Backup\Notifications\Notifications\BackupWasSuccessful::class     => ['mail'],
    //\Spatie\Backup\Notifications\Notifications\HealthyBackupWasFound::class   => ['mail'],
    //\Spatie\Backup\Notifications\Notifications\CleanupWasSuccessful::class    => ['mail'],
],
```
I get this error:
```
Exception message: There is no notification class that can handle event Spatie\Backup\Events\BackupWasSuccessful.

Exception message: There is no notification class that can handle event Spatie\Backup\Events\CleanupWasSuccessful.
```

I would prefer not to get an email on every successful backup.

freekmurze:
Do not comment those lines out in the config. Use an empty array at the end instead.

warmwhisky:
By empty array do you mean [] or ['']? its not clear in the docs how to achieve this.

freekmurze:
An empty array: `[]`.

PR #1741, opened 2023-12-01
Update overview.md

stebogit:
I was surprised [this](https://github.com/spatie/laravel-backup/issues/613#issuecomment-360397529) was never documented, it took me a bit to find out


freekmurze:
I'm going to pass on this for now. It would be better if our code would assume an empty array when no value is set.

stebogit:
Sure, but meanwhile wouldn't it be better to document the current behavior ?

freekmurze:
I feel like not to much people run into this, and it feels "heavy" to have that info there.

PR #1742, opened 2023-12-08
nunomaduro/larastan abandoned

parth391:
Abandoned warning

`Package nunomaduro/larastan is abandoned, you should avoid using it. Use larastan/larastan instead.`


freekmurze:
Thanks!

PR #1743, opened 2023-12-09
Fix translations for keys newest_backup_date and oldest_backup_date on tr language

sdkakcy:


freekmurze:
Thanks!

sdkakcy:
It's my pleasure. Thanks for package. @freekmurze

PR #1751, opened 2024-01-11
Made it possible to run the backup, cleanup and monitor commands in isolated mode

pieterclaerhout:
I updated the `backup`, `clean` and `monitor` commands so that they can be run in isolated mode.

See https://github.com/spatie/laravel-backup/discussions/1750 for the reasoning behind this.


freekmurze:
It seems like the tests are failing, could you take a look?

freekmurze:
Thanks!

PR #1752, opened 2024-01-23
Fix duplicate Portuguese translations for backup dates

ricardo-lobo:
This PR resolves an issue with duplicated Portuguese translation keys for backup date. The keys for `newest_backup_date` and `oldest_backup_date` were previously mapped to incorrect values related to backup size. 

This update ensures each key has a unique and correct translation, preventing `backupDestinationProperties()` from overwriting dates with sizes in notifications.


Nielsvanpach:
Thanks!

PR #1757, opened 2024-02-06
Add Support for Laravel 11

stefanzweifel:
Update dependency constraints to support Laravel 11 and GitHub Actions workflow to run the test suite for Laravel 11 as well.

~~Test suite currently fails for Laravel 11 installations that either use Carbon v2 or Carbon v3. (`diffForHumans()` transforms 0 seconds either to "1 second ago" or "0 seconds ago".)~~

Edit: Fixed the test suite by updating the assertions.


tests/FormatTest.php
```diff
@@ -17,7 +17,7 @@
 it('can determine the age in days', function () {
     Carbon::setTestNow(Carbon::create(2016, 1, 1)->startOfDay());
 
-    expect(Format::ageInDays(Carbon::now()))->toEqual('0.00 (1 second ago)');
+    expect(Format::ageInDays(Carbon::now()->subSeconds(5)))->toEqual('0.00 (5 seconds ago)');
```
stefanzweifel, line 20: Updated the assertion here to be compatible with Carbon v2 and v3. 
Internally v3 uses floats and no longer integers.

This led to issues when in the diff was ~0 seconds. See [test run](https://github.com/stefanzweifel/laravel-backup/actions/runs/7786191155/job/21230440647#step:7:184).

There's also a graet summary of the major changes in this comment: https://github.com/laravel/framework/pull/49764#issuecomment-1904438086

freekmurze:
Thank you!

PR #1785, opened 2024-04-29
Add config option database_dump_file_use_connection_name

jbraband:
_Code and tests are added. Documentation update was omitted because a similar config ```database_dump_file_timestamp_format``` is not documented. I'm happy to add documentation should that be required for a merge._

This PR adds a config option at ```config('backup.backup.database_dump_file_use_connection_name')``` of type boolean, defaulting to ```null``` and thus uses the database name as it always has.

This config, if set to true, uses the connection's key from ```config('database.connections')``` in the dump file name instead of the database name

**Use Case**
I have 2 connections defined to the same database, each with different db-dumper options. I have one to dump only the schema (see my PR to [spatie/db-dumper#208](https://github.com/spatie/db-dumper/pull/208)) and a second to dump the data of the database, minus the data from large tables (large as in disk space) that I can recreate.

As the package stand now, it names the dump files for these 2 connections the same based on the database name and I only get one file in the zipped archive.

With this PR, I now have both files in the zipped archive, one for each connection.

I'm happy to make any recommended modifications or additions!

> Edited for grammar and clarity

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -103,9 +103,12 @@ return [
         'database_dump_file_timestamp_format' => null,
 
         /*
-         * If specified, the database dumped file name will contain the connection name in place of the database name.
+         * The base of the dump filename, either 'database' or 'connection'
+         *
+         * If 'database' (default), the dumped filename will contain the database name.
+         * If 'connection', the dumped filename will contain the connection name.
          */
-        'database_dump_file_use_connection_name' => null,
+        'database_dump_filename_base' => 'database',
 
         /*
          * The file extension used for the database dump files.
--- a/config/backup.php
+++ b/config/backup.php
@@ -103,6 +103,11 @@ return [
         'database_dump_file_timestamp_format' => null,
 
         /*
+         * If specified, the database dumped file name will contain the connection name in place of the database name.
+         */
+        'database_dump_file_use_connection_name' => null,
+
+        /*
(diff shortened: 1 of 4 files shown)


config/backup.php
```diff
@@ -102,6 +102,11 @@
          */
         'database_dump_file_timestamp_format' => null,
 
+        /*
+         * If specified, the database dumped file name will contain the connection name in place of the database name.
+         */
+        'database_dump_file_use_connection_name' => null,
```
rcerljenko, line 108: since this is a bool value, shoud it be `false` by default instead of `null`?

freekmurze: Yeah, I agree,  `false` would be a better default

freekmurze:
Thanks for this!

Could you also add docs?

config/backup.php
jbraband: I agree. Had a thought though to change the config to a string, either 'database' or 'connection' and default it to connection.

I updated to that. It feels better that the 'database' option is visible in the config as default instead of a boolean to change behavior.

I'm happy to change it back to a boolean if that is preferred.

jbraband:
> Thanks for this!
> 
> Could you also add docs?

Docs added. I also added docs for database_dump_file_timestamp_format which were missing.

freekmurze:
Thanks!

PR #1788, opened 2024-05-02
Added null option for unlimited backup storage size

rcerljenko:
Hi,

I've added unlimited storage option for backups. We had a situation where we wanted to keep our backups for 30 days (backup command is run every hour) and it's a fairly large databse which grows each day so it's a little hard to determine max limit for storage and it feels kinda wrong to set it to some very very big number.

Thx!

### DIFF
--- a/src/Tasks/Cleanup/Strategies/DefaultStrategy.php
+++ b/src/Tasks/Cleanup/Strategies/DefaultStrategy.php
@@ -102,7 +102,7 @@ class DefaultStrategy extends CleanupStrategy
         $maximumSize = $this->config->get('backup.cleanup.default_strategy.delete_oldest_backups_when_using_more_megabytes_than')
             * 1024 * 1024;
 
-        if (($backups->size() + $this->newestBackup->sizeInBytes()) <= $maximumSize) {
+        if (!$maximumSize || ($backups->size() + $this->newestBackup->sizeInBytes()) <= $maximumSize) {
             return;
         }
         $oldest->delete();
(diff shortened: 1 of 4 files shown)


freekmurze:
Instead of `0` could you use `null` to use this feature?

rcerljenko:
> Instead of `0` could you use `null` to use this feature?

Sure

src/Tasks/Cleanup/Strategies/DefaultStrategy.php
```diff
@@ -99,10 +99,9 @@ protected function removeOldBackupsUntilUsingLessThanMaximumStorage(BackupCollec
             return;
         }
 
-        $maximumSize = $this->config->get('backup.cleanup.default_strategy.delete_oldest_backups_when_using_more_megabytes_than')
-            * 1024 * 1024;
+        $maximumSize = $this->config->get('backup.cleanup.default_strategy.delete_oldest_backups_when_using_more_megabytes_than');
 
-        if (!$maximumSize || ($backups->size() + $this->newestBackup->sizeInBytes()) <= $maximumSize) {
+        if ($maximumSize === null || ($backups->size() + $this->newestBackup->sizeInBytes()) <= $maximumSize * 1024 * 1024) {
```
freekmurze, line 104: This is getting a bit unreadable. Could you extract this check to a separate function with a good name and early returns so we can avoid using `||`

rcerljenko: No problem

freekmurze:
Thanks!

Issue #1808, opened 2024-07-02
Types v9


rcerljenko:
Hi,

This happens when you set `null` on that config in v9:

![image](https://github.com/spatie/laravel-backup/assets/16762056/7a1273ce-6fa4-49f9-ac1f-bf28499e9406)

rcerljenko:
Also, `notifications.mail.to` used to accept `array` because Laravel accepts it too in mail handler so you could send notification to multiple emails at once, but now, via config DTO, it only accepts `string`.  :( could that be changed to also accept `string|array` ?

rcerljenko:
I made a PR to fix this: https://github.com/spatie/laravel-backup/pull/1809

freekmurze:
We'll handle this further in the PR

PR #1809, opened 2024-07-02
[9.x] Fixed types

rcerljenko:
Fixed types described in more detail here: https://github.com/spatie/laravel-backup/issues/1808


Nielsvanpach:
The nullable config value is already released in 9.0.1

Could you also add a test for the array solution?

rcerljenko:
i see that there's a parallel PR for this... so maybe switch to that one instead?

Issue #1815, opened 2024-07-09
Discord Webhook not working if username is an empty string


pratikkuikel:
Steps to reproduce : 
1. Create a new webhook for a discord channel.
2. Place the webhook url in backup.php config file.
```php 
        'discord' => [
            'webhook_url' => 'https://discord.com/api/webhooks/12344',

            /*
             * If this is an empty string, the name field on the webhook will be used.
             */
            'username' => '',

            /*
             * If this is an empty string, the avatar on the webhook will be used.
             */
            'avatar_url' => '',
        ],
```
3. Change backup's notifications channel from mail to discord
```php
        'notifications' => [
            \Spatie\Backup\Notifications\Notifications\BackupHasFailedNotification::class => ['discord'],
            \Spatie\Backup\Notifications\Notifications\UnhealthyBackupWasFoundNotification::class => ['discord'],
            \Spatie\Backup\Notifications\Notifications\CleanupHasFailedNotification::class => ['discord'],
            \Spatie\Backup\Notifications\Notifications\BackupWasSuccessfulNotification::class => ['discord'],
            \Spatie\Backup\Notifications\Notifications\HealthyBackupWasFoundNotification::class => ['discord'],
            \Spatie\Backup\Notifications\Notifications\CleanupWasSuccessfulNotification::class => ['discord'],
        ],
```
4. Run `Php artisan backup:run`
5. See the console info 'Backup completed!'
6. Take a look at discord channel and it has no message notification.
7. Log response on this file : vendor\spatie\laravel-backup\src\Notifications\Channels\Discord\DiscordChannel.php
```php
    public function send($notifiable, Notification $notification): void
    {
        $discordMessage = $notification->toDiscord(); // @phpstan-ignore-line

        $discordWebhook = $notifiable->routeNotificationForDiscord();

        $response = Http::post($discordWebhook, $discordMessage->toArray());
        info($response);
    }
```
8. Run the backup command again and check laravel.log
```php
[2024-07-09 14:48:58] local.INFO: {"username": ["Must be between 1 and 80 in length.", "Username cannot be \"\""]}  
```
9. Set config file's discord username to any string and run the backup command again. Now the discord channel gets the notification.

RVxLab:
According to the [documentation](https://discord.com/developers/docs/resources/webhook#execute-webhook-jsonform-params), `username` is an optional field. So theoretically if that field (and `avatar_url` if the same rules apply) with a `!empty` or `filled` check, that should solve the issue.

PR #1817, opened 2024-07-11
[9.x] `username` parameter in `from` method should be nullable

mho22:
Based on issue #1815, `username` parameter in `from` method should be nullable in `DiscordMessage.php` on line `13` :

`src/Notifications/Channels/Discord/DiscordMessage.php` 

```php
 protected string $username = 'Laravel Backup';
 
 ...
 public function from(string $username, string $avatarUrl = null): self
 {
         $this->username = $username;
...
```

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -235,7 +235,11 @@ return [
         'discord' => [
             'webhook_url' => '',
 
-            'username' => null,
+            /*
+             * If this is an empty string, the name field will be 'Laravel Backup'.
+             * If this is set to null, the name field on the webhook will be used.
+             */
+            'username' => '',
 
             'avatar_url' => null,
         ],
--- a/config/backup.php
+++ b/config/backup.php
@@ -235,15 +235,9 @@ return [
         'discord' => [
             'webhook_url' => '',
 
-            /*
-             * If this is an empty string, the name field on the webhook will be used.
-             */
-            'username' => '',
+            'username' => null,
 
-            /*
-             * If this is an empty string, the avatar on the webhook will be used.
-             */
-            'avatar_url' => '',
+            'avatar_url' => null,
         ],
     ],
 
--- a/src/Notifications/Channels/Discord/DiscordMessage.php
+++ b/src/Notifications/Channels/Discord/DiscordMessage.php
@@ -31,10 +31,12 @@ class DiscordMessage
 
     protected string $url = '';
 
-    public function from(string $username, ?string $avatarUrl = null): self
+    public function from(?string $username = null, ?string $avatarUrl = null): self
     {
-        $this->username = $username;
-
+        if (! is_null($username)) {
+            $this->username = $username;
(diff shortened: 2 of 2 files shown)


RVxLab:
This won't actually fix the issue. The error in the original issue states that the username must be between 1 and 80 characters:

> [2024-07-09 14:48:58] local.INFO: {"username": ["Must be between 1 and 80 in length.", "Username cannot be \"\""]}  

Checking for null doesn't solve that.

I suggest that instead of changing the signature to allow nulls, don't add the `username` key to the array in the `toArray` method.

As per [the Discord documentation](https://discord.com/developers/docs/resources/webhook#execute-webhook-jsonform-params), the `username` field is optional. So doing something like

```php
public function toArray()
{
    $data = [
        'avatar_url' => $this->avatarUrl,
        'embeds' => [
            [
                'title' => $this->title,
                'url' => $this->url,
                'type' => 'rich',
                'description' => $this->description,
                'fields' => $this->fields,
                'color' => hexdec((string) $this->color),
                'footer' => [
                    'text' => $this->footer ?? '',
                ],
                'timestamp' => $this->timestamp ?? now(),
            ],
        ],
    ];

    if (!empty($this->username)) {
        $data['username'] = $this->username;
    }

    return $data;
    
}
```

Will address this point properly. When the `username` field is not passed, the default name in the webhook is sent instead.

mho22:
@RVxLab You're right, this doesn't fix the issue the way you want. I initially suggested this PR because if you don't specify `username` [ or `avatar_url` ] in the config file, an exception occurs. After reading your issue, I thought this PR might address your problem by not requiring username and using `Laravel Backup` as the default.

I've considered this further. I was torn between making `Laravel Backup` the default value for `username` and making `username` optional in `toArray()`.

Let's wait for input from a maintainer before modifying the PRs with your suggestion.

RVxLab:
> You're right, this doesn't fix the issue the way you want.

This isn't about what I want or don't want. Changing the username from `string` to `?string` and then checking for `null` doesn't solve the issue that was being reported.

If the `username` property is an empty string and passed to Discord's API like that, the call throws a validation error, which contradicts [the config option's comment](https://github.com/spatie/laravel-backup/blob/932c05ea5caee27eb07a5b8adc58c7164191a024/config/backup.php#L239).

Based on this and Discord's documentation stating that it's an optional field (analogous to `nullable|between:1,80` in Laravel's validation. Making the default value `null` and only checking for that doesn't solve the issue.

mho22:
@RVxLab I found this PR #1634 that addresses the same issue as mine. It suggests replacing an empty string with Laravel Backup, ensuring that a username is always returned. This is likely the behavior they want to maintain. We should wait for a response from @freekmurze. 

What is currently right, is the wrong comment in the config file that should be modified.

freekmurze:
Fixed by #1634

Issue #1813, opened 2024-07-08
New version need properties optional in previous package version 


mho22:
With version `9.0.0` a lot of config values become mandatory in config file. In my case in previous version `8.8.1` I didn't need these properties : 

```
backup.source.files.relative_path
backup.database_dump_filename_base
backup.database_dump_file_extension
backup.destination.compression_method
backup.destination.compression_level
backup.password
backup.encryption
backup.tries
backup.retry_delay

notifications.mail.to
notifications.slack.webhook
notifications.slack.channel
notifications.slack.username
notifications.slack.icon
```

Should this be a good idea to merge arrays with default config inside `Config::fromArray` in `src/Config/Config.php` file on line `24` : 

```php
 /** @param array<mixed> $data */
 public static function fromArray(array $data): self
 {
     $source = require realpath($raw = __DIR__.'/../config/backup.php') ?: $raw;

     return new self(
         backup: BackupConfig::fromArray(array_merge($source['backup'], $data['backup'])),
         notifications: NotificationsConfig::fromArray(array_merge($source['notifications'], $data['notifications'])),
         monitoredBackups: MonitoredBackupsConfig::fromArray(array_merge($source['monitor_backups'], $data['monitor_backups'])),
         cleanup: CleanupConfig::fromArray(array_merge($source['cleanup'], $data['cleanup']))
     );
 }
```

This is probably a bit tricky, but the idea to merge missing elements from default config into the current one should probably be the solution ?

Nielsvanpach:
I suggest to update your own config file when upgrading to a new major, so they keep in sync.

https://github.com/spatie/laravel-backup/blob/main/config/backup.php

mho22:
Hi @Nielsvanpach,

This isn't really an "issue"—it's more of a suggestion. As you know it, in Laravel 11, when developers don't need to change the default values in a config file, it's not necessary to include those defaults in the project's config.

In Laravel Backup 8 with Laravel 11, if I wanted to modify just the `backup.name`, I could do so with a minimal config file :

`config/backup.php`
```
<?php

return [

    'backup' => [ 
        
        'name' => "Foo",
        
    ],
];
```

However, in Laravel Backup 9 with Laravel 11, to modify the `backup.name`, I have to include additional mandatory elements :

`config/backup.php`
```
<?php

return [

    'backup' => [

        'name' => "Foo",

        'source' => [
           
            'files' => [

                'include' => [
                    base_path(),
                ],

                'exclude' => [
[...]

PR #1818, opened 2024-07-11
[8.x] `username` parameter in `from` method should be nullable

mho22:
Based on issue #1815, `username` parameter in `from` method should be nullable in `DiscordMessage.php` on line `13` :

`src/Notifications/Channels/Discord/DiscordMessage.php`

```php
protected string $username = 'Laravel Backup';
 
...
public function from(string $username, string $avatarUrl = null): self
{
        $this->username = $username;
...
```

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -236,14 +236,12 @@ return [
             'webhook_url' => '',
 
             /*
-             * If this is an empty string, the name field on the webhook will be used.
+             * If this is an empty string, the name field will be 'Laravel Backup'.
+             * If this is set to null, the name field on the webhook will be used.
              */
             'username' => '',
 
-            /*
-             * If this is an empty string, the avatar on the webhook will be used.
-             */
-            'avatar_url' => '',
+            'avatar_url' => null,
         ],
     ],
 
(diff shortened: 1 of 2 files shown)


Nielsvanpach:
I think this is fixed in this PR: https://github.com/spatie/laravel-backup/pull/1821

Feel free to reopen if not the case!

mho22:
@Nielsvanpach, this is applicable to version 9, not version 8. Therefore, I will need to update to version 9, but this will cause [issue #1813](#1813) to occur.

RVxLab:
Much like I mentioned in my comment in #1817, that won't fix the issue that was fixed in #1821 

@Nielsvanpach would you (or Freek) accept a PR to backport that fix to v8? I don't know what your typical stance is on older major versions.

If not, then @mho22 you would be better off forking and maintaining it yourself.

freekmurze:
> accept a PR to backport that fix to v8?

Yup, send a PR to the v8 branch and I'll tag a new release.

RVxLab:
PR has been created.

RVxLab:
@mho22 Fix for the Discord notifications has been backported. If you update to `^8.8.2` it'll start working.

PR #1825, opened 2024-07-24
Optimize storage and speed when making an encrypted backup

pellaras:
**This PR addresses two issues:**

- There is a chance of a backup to be stored as unencrypted, without noticing it.
- The free disk space needed for an encrypted backup, is double of what  an unencrypted backup needs.

**In more detail:**

Using a Listener to wait for a successful creation of backup had the side effect of first creating a zip archive, unencrypted, then create the encrypted zip archive using the unencrypted one, which demanded double the available disk space, as well as more time.

Further to the above, and most concerning, is when the Listener fails to encrypt the zip file, it shows a warning about a notification failed to be sent, but it continues to store the archive on the destination disks, allowing for the unencrypted version to make it to a remote destination with possible privacy implications. A very likely scenario for this to happen, is when there is not enough disk space available for the encrypted backup to be generated, but the unencrypted one was successfully created.

This PR tackles the issue by setting the zip archive to use encryption as soon as the files are added, without the need of first making an unencrypted version on the temp folder. This also has a nice side-effect of needing the same available disk space as if no encryption was configured.

Happy to create a PR also for v9

Note: Tide-up the tests related to the encryption, as the Listener is no longer used.


flexchar:
LGTM! Would you be as kind to spare a few minutes on this one @freekmurze?:)

freekmurze:
Will do to, could you rebase against `main` to resolve the conflicts?

pellaras:
@freekmurze are you asking me? If yes, I can do a PR against main, after getting this merged under v8.

Nielsvanpach:
There are merge conflicts preventing this from being merged. You can rebase against the v8 branch.

pellaras:
@freekmurze rebased

pellaras:
Is there anything else that can be done to help merge this PR?

I am emphasizing the issue of having unencrypted archives make it to a remote storage (i.e. s3) under specific circumstances.

src/Tasks/Backup/BackupJob.php
```diff
@@ -228,11 +228,7 @@ protected function createZipContainingEveryFileInManifest(Manifest $manifest): s
 
         consoleOutput()->info("Created zip containing {$zip->count()} files and directories. Size is {$zip->humanReadableSize()}");
 
-        if ($this->sendNotifications) {
-            $this->sendNotification(new BackupZipWasCreated($pathToZip));
-        } else {
-            app()->call('\Spatie\Backup\Listeners\EncryptBackupArchive@handle', ['event' => new BackupZipWasCreated($pathToZip)]);
-        }
+        $this->sendNotification(new BackupZipWasCreated($pathToZip));
```
freekmurze, line 231: We should only send notifications if `$this->notifications` is set to `true`

src/Tasks/Backup/Encryption.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Spatie\Backup\Tasks\Backup;
+
+class Encryption
```
freekmurze, line 5: Rename this class to `BackupEncryptionConfig`

src/Tasks/Backup/Zip.php
```diff
@@ -126,4 +135,44 @@ public function count(): int
...
+            }
+        }
+    }
+
+    public function getEncryption(): ?Encryption
+    {
+        return $this->encryption;
+    }
+
+    public function loadEncryption()
```
freekmurze, line 157: Move all of this logic to a new static method `loadFromConfigFile` on the `BackupEncryptionClass` so the `Zip` becomes smaller

freekmurze (changes requested):

freekmurze:
Closing due to inactivity

PR #1821, opened 2024-07-14
[9.x] Fix empty string in `backup.notifications.discord.username` to fail sending a notification

RVxLab:
Fixes #1815 

This PR fixes an issue caused by using an empty string for the Discord Webhook `username` property.

According to [Discord's documentation](https://discord.com/developers/docs/resources/webhook#execute-webhook-jsonform-params), the `username` field is optional and when set, requires it to be between 1 and 80 characters long. This lines up with the `name` field when [creating a webhook](https://discord.com/developers/docs/resources/webhook#create-webhook).

The current implementation of the notification doesn't follow that. If the `backup.notifications.discord.username` is an empty string (which is currently the default), the notification is never sent due to a validation error. See the mentioned issue above for details.

This PR fixes these issues by only setting the username set in the config if it's not empty.

To test this, I created a server and added a webhook with the name Test McWebhook:

image

I applied the fix to a project locally and ran it with 3 usernames in order:

- `'Laravel Backup'`
- `''`
- `'Some Other Custom Name'`

image

As a final change, I removed the `?? 'Laravel Backup'` expression when setting the username. Since `$username` is non-nullable, that expression is never run.

PR #1829, opened 2024-08-07
[8.x] Only add the username property if it's not empty

RVxLab:
Backport of #1821 for the v8 branch

See also #1818


freekmurze:
Thanks!

PR #1838, opened 2024-09-05
Add customize the locale to specify the backup locale for translation purposes

junaidiar19:
This PR is just for create specify the backup locale for custom translation.

Here is the reason I address this PR:
- Our base application used locale ID but we want backup notifications using English.
- I have not found in this package how to customize the locale.

That's all for me Happy Coding.

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -338,4 +338,10 @@ return [
         'retry_delay' => 0,
     ],
 
+    /**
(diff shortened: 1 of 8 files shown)


freekmurze:
I'm going to pass on this to keep the package simple.

To get the backup notifications in a specific language, you could extend the default ones and set the language you want in your extended versions. Don't forget to register your custom notifications classes in the config file.

Issue #1840, opened 2024-09-17
mysqldump: Got error: 2026: "TLS/SSL error: self-signed certificate"


diadal:
All services that utilize DB are working fine except `laravel-backup`

Exception message: The dump process failed with a none successful exitcode. Exitcode ======== 2: Misuse of shell builtins Output ====== Error Output ============ mysqldump: Got error: 2026: "TLS/SSL error: self-signed certificate" when trying to connect

[...]

PR #1841, opened 2024-09-17
Update DbDumperFactory.php

diadal:
fix for spatie/laravel-backup#1840

ability to use skip SSL

### DIFF
--- a/src/Tasks/Backup/DbDumperFactory.php
+++ b/src/Tasks/Backup/DbDumperFactory.php
@@ -47,6 +47,7 @@ class DbDumperFactory
 
         if ($dbDumper instanceof MySql) {
             $dbDumper
+                ->setSkipSsl($dbConfig['dump']['skipSsl'] ?? false)
                 ->setDefaultCharacterSet($dbConfig['charset'] ?? '')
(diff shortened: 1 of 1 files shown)


src/Tasks/Backup/DbDumperFactory.php
```diff
@@ -47,6 +47,7 @@ public static function createFromConnection(string $dbConnectionName): DbDumper
 
         if ($dbDumper instanceof MySql) {
             $dbDumper
+                ->setSkipSsl($dbConfig['dump']['skipSsl'] ?? false)
```
riasvdv, line 50: Can you use `skip_ssl` instead as the other config options use underscores, can you also add this to the docs?

riasvdv (changes requested):

src/Tasks/Backup/DbDumperFactory.php
diadal: I will this weekend

diadal: done, also remember to merge this https://github.com/spatie/db-dumper/pull/213 if not it won't work

PR #1846, opened 2024-09-29
Allow overriding default implementation of TemporaryDirectory

jonerickson:
This PR arises out of the need to change the default implementation of the `TemporaryDirectory` service. In our story, serving a multi-tenant application and using this package to backup each individual tenant's DB, we needed the ability to customize the temporary directory's location. Because we have a large number of tenants, we elected to create a batch job to backup each tenant's DB. Because of this, we are running into the issue where multiple jobs are trying to use the same temporary directory. As you can imagine, this creates chaos. We now have the ability to set the file path based on the current tenant context allowing each tenant to use their own temporary directory.

Extending the implementation:
```
$this->app->extend(TemporaryDirectory::class, function (TemporaryDirectoryAdapter $temporaryDirectory) {
    return new TenantTemporaryDirectory($temporaryDirectory);
});
```

Customizing our tmp path:
```
class TenantTemporaryDirectory extends TemporaryDirectory implements TemporaryDirectoryContract
{
    public function __construct(TemporaryDirectory $temporaryDirectory)
    {
        parent::__construct($temporaryDirectory->location);
    }

    protected function getFullPath(): string
    {
        if (! tenancy()->initialized) {
            return parent::getFullPath();
        }

        $tenantId = tenant()->getTenantKey();

        return storage_path(optional($tenantId, fn ($id) => "app/backup-temp/tenant$id") ?? 'app/backup-temp').DIRECTORY_SEPARATOR.$this->name;
    }
}
```


jonerickson:
Hey Spatie team, seeing if you have any questions regarding this PR. Thanks!

src/BackupServiceProvider.php
```diff
@@ -49,6 +52,7 @@ public function packageRegistered(): void
         $this->app->singleton(ConsoleOutput::class);
 
         $this->app->bind(CleanupStrategy::class, config('backup.cleanup.strategy'));
+        $this->app->bind(TemporaryDirectoryContract::class, fn() => new TemporaryDirectoryAdapter(config('backup.temporary_directory') ?? storage_path('app/backup-temp')));
```
freekmurze, line 55: Added a contract and adapter feels a bit like overkill to me. Maybe just bind the `backup-temporary-project` string to a function that returns a new instance of temporary directory directly.

In your project you could just change the binding to whatever you want.

freekmurze (changes requested):

src/BackupServiceProvider.php
jonerickson: Thanks for the feedback. I have removed both the contract and the adapter and just bound the implementation to a string.

PR #1872, opened 2025-01-22
Fix temp directory config path and add unit test

### LINKED ISSUE #1871: temporary_directory config broken
https://github.com/spatie/laravel-backup/pull/1846 This pull request changed the config path from backup.backup.temporary_directory to backup.temporary_directory.
So since patch 9.2.1 all configs including the default config in github are broken.
In our system this is breaking, because we use a read only docker container.

NiroDeveloper:
Fix #1871 this bug was created by #1846


freekmurze:
Thank you!

PR #1873, opened 2025-01-23
[BUG] merge published config with package config

pxpm:
There is currently a bug on the way the configs are merged. If a "nested array key" is missing on the published config file, `array_merge()` wouldn't add that key and the application will fail to properly boot the config. 

I got the following error after updating from v8 to v9 - ` Undefined array key "compression_method"`.

I dug a bit on it, and indeed my published `backup.php` config file didn't had that key, but that key is present on the package config file: https://github.com/spatie/laravel-backup/blob/fe980cee0a5d2b0773395da3446cd81a3288100a/config/backup.php#L133 so I would expect the "default package value" to be used. 

This PR changes `array_merge` to `array_replace_recursive`, so that we use the package config as the base, and replace any values from the published config file. 

I've added tests for both scenarios (without keys present in the published config file, and with keys present in the config file). 

@freekmurze any chance you can take a look at this ? 

Cheers

### DIFF
--- a/src/Config/Config.php
+++ b/src/Config/Config.php
@@ -27,10 +27,10 @@ class Config extends Data
         $source = require dirname(__DIR__, 2).'/config/backup.php';
 
         return new self(
-            backup: BackupConfig::fromArray(array_merge($source['backup'], $data['backup'] ?? [])),
-            notifications: NotificationsConfig::fromArray(array_merge($source['notifications'], $data['notifications'] ?? [])),
+            backup: BackupConfig::fromArray(array_replace_recursive($source['backup'], $data['backup'] ?? [])),
+            notifications: NotificationsConfig::fromArray(array_replace_recursive($source['notifications'], $data['notifications'] ?? [])),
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

cannahan:
After this update, disabling notifications in the backup.php configuration file using the following line is no longer possible: “\Spatie\Backup\Notifications\Notifications\BackupWasSuccessfulNotification::class => [],” because the empty array is not recognized as an actual value by the array_replace_recursive() function, but as an additional array.

mho22:
@pxpm This is why I didn't use `array_replace_recursive` in my previous pull request. You can't overwrite array with data with an empty array using `array_replace_recursive` like @cannahan mentionned : 


> `array_replace_recursive()` replaces the values of array with the same values from all the following arrays. If a key from the first array exists in the second array, its value will be replaced by the value from the second array. If the key exists in the second array, and not the first, it will be created in the first array. `If a key only exists in the first array, it will be left as is`. If several arrays are passed for replacement, they will be processed in order, the later array overwriting the previous values.


The inconvenience of my PR is that you stil have to add some nested array keys even if you don't need them but overall most of them are no more needed.

PR #1875, opened 2025-01-27
remove the `only()` from tests

pxpm:
Sorry @freekmurze I forgot to remove the `only()` while writing the tests for PR #1873 .

🙏


freekmurze:
Oh, didn't catch that one myself. Thanks for the quick PR! :thumbs

Issue #1877, opened 2025-01-29
Command not found: `mysqldump` error on running backup with empty databases list


alies-dev:
Such error happens on running `backup:run` when `databases` array in the config is empty:
```php
// backup.php
return [
    'backup' => [
        'source' => [
            'databases' => [], // empty list
        ],
    ],
];
```

```
The dump process failed with a none successful exitcode.
Exitcode
========
127: Command not found

Output
======
<no output>

Error Output
============
sh: 1: mysqldump: not found

Context:

[
  {
    "exception": "[object] (Spatie\\Backup\\Exceptions\\BackupFailed(code: 0): The dump process failed with a none successful exitcode.
Exitcode
========
127: Command not found

Output
======
<no output>

Error Output
============
sh: 1: mysqldump: not found
 at %project%/vendor/spatie/laravel-backup/src/Exceptions/BackupFailed.php:17)
[stacktrace]
#0 %project%/vendor/spatie/laravel-backup/src/Tasks/Backup/BackupJob.php(187): Spatie\\Backup\\Exceptions\\BackupFailed::from()
#1 %project%/vendor/spatie/laravel-backup/src/Commands/BackupCommand.php(72): Spatie\\Backup\\Tasks\\Backup\\BackupJob->run()
#2 %project%/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(36): Spatie\\Backup\\Commands\\BackupCommand->handle()
#3 %project%/vendor/laravel/framework/src/Illuminate/Container/Util.php(43): Illuminate\\Container\\BoundMethod::{closure:Illuminate\\Container\\BoundMethod::call():35}()
#4 %project%/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(95): Illuminate\\Container\\Util::unwrapIfClosure()
#5 %project%/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(35): Illuminate\\Container\\BoundMethod::callBoundMethod()
#6 %project%/vendor/laravel/framework/src/Illuminate/Container/Container.php(694): Illuminate\\Container\\BoundMethod::call()
#7 %project%/vendor/laravel/framework/src/Illuminate/Console/Command.php(213): Illuminate\\Container\\Container->call()
#8 %project%/vendor/symfony/console/Command/Command.php(279): Illuminate\\Console\\Command->execute()
[...]

PR #1878, opened 2025-01-29
Add a test for the empty DB list

alies-dev:
To prove #1877 issue (it's not a fix, just a failing test)


freekmurze:
I've manually added the test in our test suite and it seems to pass...

Issue #1879, opened 2025-01-30
Mail notification is sent when it's not supposed to


rcerljenko:
Hi,

Since latest update (9.2.4) system is sending mail notification even if it doesn't suppose to.

Given this config:
![Image](https://github.com/user-attachments/assets/024ab183-87da-43c3-a71d-8d090be7c912)

`CleanupWasSuccessfulNotification` shouldn't be sent but it is. I'm suspecting that it has something to do with the latest PR... specifically this line:

https://github.com/spatie/laravel-backup/pull/1873/files#diff-18d9e9c6768a133f4fc7723d566dc36b5a1a07bee3c30396619c76cd85422bfeR31

Okipa:
I can confirm this erratic behavior on my end 👍

freekmurze:
@pxpm Could you take a look at this? If we don't find a fix for this soonish, I'll revert the PR.

pxpm:
On it 👍

julesjanssen:
The same applies to the `source.files` part of the config. It's no longer possible to leave that array empty.

freekmurze:
Fixed in the latest release.

rcerljenko:
many thanks guys!

Okipa:
Thanks!

PR #1880, opened 2025-01-30
account for empty array in notifications config

pxpm:
reported in https://github.com/spatie/laravel-backup/issues/1879

Given how the configurations are built, `array_replace_recursive` won't allow to use an empty array as replacement. 

I am not 100% happy with this solution as it requires to treat `$notifications` array differently from other configs. So feel free to close this and revert the other PR. 

Anyway I added tests for this changes to ensure it's working as expected, I just don't like to have that "special" treatment of notifications array. 

This wouldn't be an issue if instead of an empty array, the way to disable the notification was setting it to `null/false`, but at this moment it would be a breaking change. 

@freekmurze

### DIFF
--- a/src/Config/Config.php
+++ b/src/Config/Config.php
@@ -26,9 +26,11 @@ class Config extends Data
     {
         $source = require dirname(__DIR__, 2).'/config/backup.php';
 
+        $data['notifications']['notifications'] = array_replace($source['notifications']['notifications'], $data['notifications']['notifications'] ?? []);
+
         return new self(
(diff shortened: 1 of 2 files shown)


freekmurze:
I'm going to revert the original PR for now, so we don't need to have special cases in our code.

pxpm:
howww, missed it for a minute!

I've just updated this branch to remove the "special" case, and created a seamless way to merge the configs that should work no matter how deep they are. 😢 

I opened the PR at: https://github.com/spatie/laravel-backup/pull/1881

PR #1881, opened 2025-01-31
add config merger class

pxpm:
This class will correctly handle merging configs. 

None of the native php functions handle the merge process of the config arrays given the current setup. 

This class ensures arrays are merged correctly and "empty" arrays are respected.

This is a super-seed of #1880 that was reverted. 👍

### DIFF
--- a/src/Config/Config.php
+++ b/src/Config/Config.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\Backup\Config;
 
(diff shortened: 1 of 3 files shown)


freekmurze:
I'm going to pass on this for now. To avoid problems, with merging I'll suggest keeping the published config file updated with the structure of the latest one in the package.

Issue #628, opened 2018-02-10
Type error: Argument 1 passed to Spatie\DbDumper\DbDumper::setPort() must be of the type integer, string given


gauravmak:
I am getting the following error while trying to run _php artisan backup:run_
What may be the cause?

![screenshot from 2018-02-10 19-57-49](https://user-images.githubusercontent.com/11887260/36063004-fb49d58c-0e9c-11e8-8b99-ae2e9ab658cd.png)

freekmurze:
What does you database config look like?

gauravmak:
It works with a fresh Laravel installation. So, there's something specific to this project only.

![screenshot from 2018-02-10 21-36-36](https://user-images.githubusercontent.com/11887260/36064015-b9faa378-0eaa-11e8-8e55-f25e6d71c48a.png)

freekmurze:
Could you try setting to value of port to `3306` instead of `'3306'`?

gauravmak:
No luck with that. New installation also has a string port number and it works.
Knowing that this has to do something with my project, I won't take more of your time; it's very valuable. I will try my best to troubleshoot myself first and update here.

Thank, Freek. You're awesome. Looking forward to seeing you in person at Laravellive this year.

gauravmak:
The issue was with my .env file. DB_PORT was set to be blank. Not sure when I did it. Sorry, for that. Thanks.

shazaman23:
I'm seeing this error, but my .env file and config/database.php file both look fine. 
*.env*
`DB_PORT=3306`

*config/database.php*
```
'mysql' => [
            'driver' => 'mysql',
            'host' => env('DB_HOST', '127.0.0.1'),
            'port' => env('DB_PORT', '3306'),
            'unix_socket' => env('DB_SOCKET', ''),
            'database' => env('DB_DATABASE', 'killfood'),
            'username' => env('DB_USERNAME', 'homestead'),
            'password' => env('DB_PASSWORD', 'secret'),
            'charset' => 'utf8mb4',
            'collation' => 'utf8mb4_unicode_ci',
            'prefix' => '',
            'strict' => true,
            'engine' => null,
            'modes'  => [
                'ONLY_FULL_GROUP_BY',
                'STRICT_TRANS_TABLES',
                'NO_ZERO_IN_DATE',
                'NO_ZERO_DATE',
                'ERROR_FOR_DIVISION_BY_ZERO',
                'NO_ENGINE_SUBSTITUTION',
            ],
[...]

PR #1901, opened 2025-04-20
DbDumperFactory: Ignore invalid ports

CloCkWeRX:
When an environment variable is set to blank, this is often translated to `''` rather than `null`; which leads immediately to 
```
Type error: Argument 1 passed to Spatie\DbDumper\DbDumper::setPort() must be of the type integer, string given
```

Fixes https://github.com/grokability/snipe-it/issues/14374
Fixes https://github.com/spatie/laravel-backup/issues/628

Other things to think about:

- This should warn when it ignores input; just wasn't sure about what the laravel way to do this was
- Other things that are using `isset()` but could be `!empty()`

### DIFF
--- a/src/Tasks/Backup/DbDumperFactory.php
+++ b/src/Tasks/Backup/DbDumperFactory.php
@@ -57,7 +57,9 @@ class DbDumperFactory
         }
 
         if (isset($dbConfig['port'])) {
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #1903, opened 2025-04-23
Use MariaDb dumper

NiroDeveloper:
The spatie/laravel-backup package currently uses MySQL dumper for all MySQL-compatible databases, including MariaDB.

However, MariaDB has deprecated the mysqldump utility in favor of mariadb-dump as documented in their official documentation.

Recently, a dedicated MariaDB dumper was added to the spatie/db-dumper package in https://github.com/spatie/db-dumper/pull/220, which properly uses the mariadb-dump utility.

This PR updates this package to utilize the new dumper when a MariaDB connection is required.

on-behalf-of: @e-solutions-GmbH <info@esolutions.de>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "illuminate/notifications": "^10.10.0|^11.0|^12.0",
         "illuminate/support": "^10.10.0|^11.0|^12.0",
         "league/flysystem": "^3.0",
-        "spatie/db-dumper": "^3.7",
+        "spatie/db-dumper": "^3.8",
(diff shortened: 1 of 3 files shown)


freekmurze:
Nice, thanks!

chosten:
This is a breaking change.
Is there a way to use the mysql dump utility instead ? There is no mariadb dump utility on my servers.

NiroDeveloper:
Oh true, this was a bit hard for a minor change.
A workaround would be to create a symlink on the server.
Consider that using mysql-dump with mariadb is deprecated.

PR #1904, opened 2025-04-25
Improve port check condition

miskith:
Condition fails for cases when port is valid number, but defined as string as well as checks for range from 1 to 65535.


freekmurze:
Thanks!

PR #1905, opened 2025-05-12
Runtime reload configuration

inserve-paul:
See discussion: https://github.com/spatie/laravel-backup/discussions/1900

Added an option to reload the configuration. For e.g. when used in a multi tenancy environment where the command is called from another command and the config has been changed for the specific tenant.

### DIFF
--- a/docs/taking-backups/overview.md
+++ b/docs/taking-backups/overview.md
@@ -105,22 +105,30 @@ The specified databases will be dumped and, together with the selected files, zi
  
 The more files you need to backup, the bigger the zip will become. Make sure there's enough free space on your disk to create the zip file. After the source zip file has been copied to all destinations, it will be deleted.
 
-If you want to backup with a specific configuration (in this example 'backup'), or reload it runtime (for e.g. when you make runtime changes for multiple tenants backup), run:
+### Running backups with a specific configuration
+If you want to back up different areas of your Laravel application separately – for example with different schedules, database connections, filesystem disks, or cleanup settings – you can create custom backup configuration files.
+
+#### Example:
+Additional config files placed in the config/ directory:
+
+- config/backup_database.php
+- config/backup_invoices.php
+- config/backup_uploads.php
+
+You can then run backups and cleanup commands individually:
 
 ```bash
-php artisan backup:run --config=backup
+php artisan backup:run --config=backup_database
+php artisan backup:clean --config=backup_database
+
+php artisan backup:run --config=backup_invoices
+php artisan backup:clean --config=backup_invoices
+
+php artisan backup:run --config=backup_uploads
+php artisan backup:clean --config=backup_uploads
 ```
 
-Possible situations where you might want to do this are for e.g., when you want to backup multiple tenants at the same time from a console command:
-```php
-
-$tenants = ['tenant1', 'tenant2'];
-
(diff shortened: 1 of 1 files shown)


inserve-paul:
@freekmurze Would you like to take a look at this PR? Thank you!

steffensatrapa:
Thanks for your PR — it pointed me in the right direction. 
I have several backup configurations that I used to load dynamically in a BackupJob. However, after upgrading to v9, this no longer worked.

Here are a couple of suggestions:

1) Extend the --reload-config=<configname> option — this is what I’m using now to load specific configurations:

``` 
protected $signature = 'configurable-backup:run {--filename=} {--only-db} {--db-name=*} {--only-files} {--only-to-disk=} {--disable-notifications} {--timeout=} {--tries=} {--reload-config=}';
...
public function handle(): int
{
    if ($this->option('reload-config')) {
        $this->config = Config::fromArray(config($this->option('reload-config') ?? 'backup'));
    }
     ...
}
```
2) Don’t forget to extend the other backup commands as well, like backup:clean, which I also had to modify.

inserve-paul:
> Could you add a test to prove that the option works. Could you also update the docs?

@freekmurze I've updated the PR, added documentation and a test

docs/taking-backups/overview.md
```diff
@@ -104,7 +104,24 @@ This section of the configuration determines which files and databases will be b
...
+```
+
+Possible situations where you might want to do this are for e.g., when you want to backup multiple tenants at the same time from a console command:
+```php
+
+$tenants = ['tenant1', 'tenant2'];
+
+foreach ($tenants as $tenant) {
+    config(['backup.backup.destination.filename_prefix' => $tenant.'_']);
+    $this->call('backup:run', ['--timeout' => 120, '--config' => 'backup']);
```
freekmurze, line 121: I'm afraid I don't understand the example fully. 

I was expecting that you could use another back config with something like this

php artisan backup:run --config=tenant1

steffensatrapa: maybe better (my use-case):

### Running Backups with Specific Configuration

If you want to back up different areas of your Laravel application separately – for example with different schedules, database connections, filesystem disks, or cleanup settings – you can create custom backup configuration files.

#### Example:

Additional config files placed in the `config/` directory:

- `config/backup_database.php`
- `config/backup_invoices.php`
- `config/backup_uploads.php`

You can then run backups and cleanup commands individually:

```bash
php artisan backup:run --config=backup_database
php artisan backup:clean --config=backup_database

php artisan backup:run --config=backup_invoices
php artisan backup:clean --config=backup_invoices

php artisan backup:run --config=backup_uploads
php artisan backup:clean --config=backup_uploads
```
This allows full flexibility in scheduling, retention, and target destinations for each backup scope.

inserve-paul: @steffensatrapa Thanks for your addition!

@freekmurze Is that example clear? In that case I will add it to the documentation

zfm-andreasmatuszewski: Any updates here? @freekmurze

websanova: Just came across this and need this as well, the `--config` option looks great to me, when is this getting merged?

PR #1906, opened 2025-05-20
Correct danish translation for "application"

mathiaskastrup:
While "Ansøgning" in certain contexts can refer to "application", it typically suggests formal application (e.g., "jobansøgning"). In system messages, the technical term "applikation" should be employed. 

This makes the change consistent with the rest of translations.


freekmurze:
Thanks!

PR #1922, opened 2025-10-15
Add realpath error handling, temp cleanup on fatal errors, credentials sanitization

imhayatunnabi:
- Missing Error Handling for realpath() Failures
- Temporary Directory Not Cleaned Up on Certain Failures
- Database Connection Error Messages Expose Credentials
- Port Validation Accepts Invalid Ports


freekmurze:
could you rebase this with `main`?

src/Tasks/Backup/BackupJob.php
```diff
@@ -302,7 +314,8 @@ protected function copyToBackupDestinations(string $path): void
             ->each(function (BackupDestination $backupDestination) use ($path) {
                 try {
                     if (! $backupDestination->isReachable()) {
-                        throw new Exception("Could not connect to disk {$backupDestination->diskName()} because: {$backupDestination->connectionError()}");
+                        $sanitizedError = CredentialSanitizer::sanitizeMessage($backupDestination->connectionError());
```
freekmurze: I'm ok with throwing an exception here. You don't need to sanitize the message.

src/Tasks/Backup/BackupJob.php
```diff
@@ -313,7 +326,8 @@ protected function copyToBackupDestinations(string $path): void
 
                     $this->sendNotification(new BackupWasSuccessful($backupDestination));
                 } catch (Exception $exception) {
-                    consoleOutput()->error("Copying zip failed because: {$exception->getMessage()}.");
+                    $sanitizedError = CredentialSanitizer::sanitizeMessage($exception->getMessage());
```
freekmurze: Remove this, I think the current exception is good enough

src/Tasks/Backup/BackupJob.php
freekmurze: Do refactor the exception to how it's done for the other exceptions. See `BackupFailed`

src/Tasks/Backup/DbDumperFactory.php
```diff
@@ -58,13 +58,17 @@ public static function createFromConnection(string $dbConnectionName): DbDumper
         }
 
         if (isset($dbConfig['port'])) {
-            if (filter_var($dbConfig['port'], FILTER_VALIDATE_INT, [
+            $port = $dbConfig['port'];
```
freekmurze: Remove this. It'll fail elsewhere if the port is not correct.

freekmurze (changes requested):
Thank you for this. I think the changes are probably a bit too strict (and the code is a bit too complicated to my eyes).

There is one new exception that you introduced that we could keep, but could you refactor it so it follows the patterns of the other exceptions?

Issue #1932, opened 2025-11-05
Fix: allow backups to continue when some destinations fail instead of failing completely


imhayatunnabi:
An improvement is necessary to  allow backups to continue when some destinations fail instead of failing completely. For example, if we have backups going to both local and S3, and S3 is down, the backup can still succeed by saving to local storage only

PR #1933, opened 2025-11-05
Fix: allows backups to continue when some destinations fail instead of failing completely

imhayatunnabi:
When trying to use the `continue_on_failure` configuration option, the application crashes with "Unknown named parameter $continueOnFailure" error. This happens because the DestinationConfig constructor is missing the parameter that the `fromArray()` method tries to pass to it.

Allows backups to continue when some destinations fail instead of failing completely 

 https://github.com/spatie/laravel-backup/issues/1932


freekmurze:
Thanks!

PR #1978, opened 2026-06-09
Encrypt archives while they are being built

pellaras:
Currently archives are encrypted after they're created: the zip is built and closed, then the `EncryptBackupArchive` listener re-opens it and rewrites every entry with encryption. For a large backup that means writing the whole archive to disk twice.

This moves the encryption into the build step. Each file is marked for encryption as it's added to the zip, so the archive only gets written once. Nothing changes from a user's perspective: it's still driven by the `backup.backup.password` and `backup.backup.encryption` config and produces the same encrypted archive.

I kept the old listener for backwards compatibility. `EncryptBackupArchive` is now `@deprecated` and no longer registered, but it's left in place so anything that references or extends it keeps working. The `BackupZipWasCreated` event still fires.

The encryption cases (no password, encryption disabled, the AES variants, and that the archive can't be extracted without the password) are covered in `BackupCommandTest`.


freekmurze (dismissed):
Thanks, this is a nice improvement. Encrypting during the build instead of rewriting the whole archive afterwards is clearly the better approach, and the test coverage for the config matrix (no password, `none`, the AES variants, extraction without password) is good. I ran the full suite locally (155 passed) and PHPStan is clean.

One real issue and a couple of smaller points:

**`Zip::add()` now fatally crashes when called without a `$nameInZip` while a password is configured**

In `Zip::add()`, `$fileNameInZip = ltrim((string) $nameInZip, DIRECTORY_SEPARATOR)` becomes `''` when `$nameInZip` is `null` (which is also forced for every file when an array is passed). `ZipArchive::addFile($file, '')` is fine (falls back to the file path as entry name), but the new call then throws:

```
ValueError: ZipArchive::setEncryptionName(): Argument #1 ($name) must not be empty
```

The package's own flow always passes a name via `createForManifest()`, but `Zip` and `add()` are public API, so anyone using them directly with a backup password configured now gets a fatal error. Note the existing `setCompressionName()` call has a related quirk in this case (it uses `ltrim($nameInZip ?: $file, ...)`, which doesn't match the actual entry name either and silently returns `false`), so the cleanest fix is to compute one real entry name and use it everywhere:

```php
$fileNameInZip = ltrim($nameInZip ?: $file, DIRECTORY_SEPARATOR);

$this->zipFile->addFile($file, $fileNameInZip);

$this->zipFile->setCompressionName($fileNameInZip, $compressionMethod, $compressionLevel);

if ($this->encryptionAlgorithm !== null) {
    $this->zipFile->setEncryptionName($fileNameInZip, $this->encryptionAlgorithm);
}
```

That also fixes the pre-existing silent `setCompressionName` mismatch. Since encryption is a security guarantee, please also consider throwing if `setEncryptionName()` returns `false` rather than letting an entry silently end up unencrypted.

**Deprecated listener is now untested**

`EncryptBackupArchive` is kept for backwards compatibility but its test file was deleted, so the code we keep for BC is the only code with zero coverage. If it stays, I'd keep a minimal version of `EncryptBackupArchiveTest` until the listener is removed in the next major.

**Docs/changelog note for self-registered listeners**

The old docs said "You are free to add this listener your own or override it." Anyone who registered `EncryptBackupArchive` themselves will now build an encrypted zip and then have their listener rewrite every entry again, which is exactly the double write this PR removes. Worth a sentence in the docs or changelog telling those users to drop their manual registration.

Minor: in `open()`, the `($password = $this->config->backup->password) !== null` check is redundant since `$encryptionAlgorithm` is only ever set when the password is non-null.

src/Tasks/Backup/Zip.php
```diff
@@ -111,13 +121,19 @@ public function add(string|iterable $files, ?string $nameInZip = null): self
...
 
                 $this->zipFile->setCompressionName(
                     ltrim($nameInZip ?: $file, DIRECTORY_SEPARATOR),
                     $compressionMethod,
                     $compressionLevel
                 );
+
+                if ($this->encryptionAlgorithm !== null) {
+                    $this->zipFile->setEncryptionName($fileNameInZip, $this->encryptionAlgorithm);
+                }
```
freekmurze: This is the crash from my earlier review: when `add()` is called without a `$nameInZip` (or with an array, which forces it to `null`), `$fileNameInZip` becomes `''` and `setEncryptionName('')` throws `ValueError: Argument #1 ($name) must not be empty`. Computing one entry name with the same fallback the compression call already uses fixes that, and also fixes the pre-existing silent `setCompressionName()` mismatch in the no-name case. Throwing when `setEncryptionName()` returns `false` (libzip without AES support) avoids silently producing an unencrypted archive the user believes is encrypted.

```suggestion
                $fileNameInZip = ltrim($nameInZip ?: $file, DIRECTORY_SEPARATOR);

                $this->zipFile->addFile($file, $fileNameInZip);

                $this->zipFile->setCompressionName($fileNameInZip, $compressionMethod, $compressionLevel);

                if ($this->encryptionAlgorithm !== null) {
                    $result = $this->zipFile->setEncryptionName($fileNameInZip, $this->encryptionAlgorithm);

                    if ($result !== true) {
                        throw BackupFailed::from(new \Exception("Failed to set encryption for '{$fileNameInZip}' in zip file at '{$this->pathToZip}'."));
                    }
                }
```

src/Tasks/Backup/Zip.php
```diff
@@ -15,11 +15,17 @@ class Zip
...
+    protected ?int $encryptionAlgorithm = null;
+
     public function __construct(protected string $pathToZip)
     {
         $this->zipFile = new ZipArchive;
         $this->config = app(Config::class);
 
+        if ($this->config->backup->password !== null && $this->config->backup->encryption->shouldEncrypt()) {
+            $this->encryptionAlgorithm = $this->config->backup->encryption->algorithm();
+        }
```
freekmurze: `Encryption::shouldEncrypt()` is literally `$this->algorithm() !== null`, so this evaluates the match twice. Assigning `algorithm()` directly behaves identically (it returns `null` for `none`, leaving the property `null`):

```suggestion
        if ($this->config->backup->password !== null) {
            $this->encryptionAlgorithm = $this->config->backup->encryption->algorithm();
        }
```

src/Tasks/Backup/Zip.php
```diff
@@ -85,6 +91,10 @@ public function open(): void
         if ($result !== true) {
             throw BackupFailed::from(new \Exception("Failed to open zip file at '{$this->pathToZip}'. ZipArchive error code: {$result}"));
         }
+
+        if ($this->encryptionAlgorithm !== null && ($password = $this->config->backup->password) !== null) {
+            $this->zipFile->setPassword($password);
+        }
```
freekmurze: Pulling the assignment out of the condition reads better (house style avoids assignments inside compound conditions):

```suggestion
        $password = $this->config->backup->password;

        if ($this->encryptionAlgorithm !== null && $password !== null) {
            $this->zipFile->setPassword($password);
        }
```

tests/Commands/BackupCommandTest.php
```diff
@@ -384,6 +379,75 @@
...
+
+    $this->artisan('backup:run --only-db --db-name=db1 --only-to-disk=local')->assertExitCode(0);
+
+    $zip = new ZipArchive;
+    $zip->open(Storage::disk('local')->path($this->expectedZipPath));
+
+    expect($zip->statIndex(0)['encryption_method'])->toBe(ZipArchive::EM_NONE);
+
+    $zip->close();
+});
```
freekmurze: The deleted listener test also covered `encryption => null` and `encryption => false`, which `BackupConfig::parseEncryption()` explicitly handles. A dataset keeps that coverage:

```suggestion
it('does not encrypt the backup when encryption is disabled', function (mixed $encryption) {
    config()->set('backup.backup.password', '24dsjF6BPjWgUfTu');
    config()->set('backup.backup.encryption', $encryption);
    config()->set('backup.backup.source.databases', ['db1']);

    $this->artisan('backup:run --only-db --db-name=db1 --only-to-disk=local')->assertExitCode(0);

    $zip = new ZipArchive;
    $zip->open(Storage::disk('local')->path($this->expectedZipPath));

    expect($zip->statIndex(0)['encryption_method'])->toBe(ZipArchive::EM_NONE);

    $zip->close();
})->with([
    'none',
    null,
    false,
]);
```

tests/Commands/BackupCommandTest.php
```diff
@@ -384,6 +379,75 @@
...
+    $zip->open(Storage::disk('local')->path($this->expectedZipPath));
+
+    $extractionPath = $this->getTempDirectory().'/extraction';
+
+    expect($zip->extractTo($extractionPath))->toBeFalse();
+
+    $zip->setPassword($password);
+    expect($zip->extractTo($extractionPath))->toBeTrue();
+
+    $zip->close();
```
freekmurze, line 456: The deleted listener test verified the decrypted contents; here only `extractTo()`'s return value is asserted. Checking the extracted file makes the round trip explicit:

```suggestion
    $zip->setPassword($password);
    expect($zip->extractTo($extractionPath))->toBeTrue();

    $extractedFile = $extractionPath.'/'.$zip->statIndex(0)['name'];
    expect(file_exists($extractedFile))->toBeTrue();
    expect(filesize($extractedFile))->toBeGreaterThan(0);

    $zip->close();
});
```

freekmurze (commented):
Following up on my earlier review with concrete, one-click-applicable suggestions. I applied all of these locally on top of your branch and verified them: 158 tests pass (2,376 assertions) and PHPStan reports no errors.

One more thing that can't be an inline suggestion because `tests/ZipTest.php` isn't part of this diff: a regression test for the `add()`-without-name case. Note the `forgetInstance()` call, which is needed because the `beforeEach` in that file already resolves the scoped `Config` singleton before the test can set the password:

```php
it('encrypts files added without a name in zip when a password is configured', function () {
    config()->set('backup.backup.password', '24dsjF6BPjWgUfTu');
    app()->forgetInstance(\Spatie\Backup\Config\Config::class);

    $pathToZip = "{$this->getTempDirectory()}/encrypted-test.zip";

    $zip = new Zip($pathToZip);
    $zip->add(__FILE__);
    $zip->close();

    $zipArchive = new ZipArchive;
    $zipArchive->open($pathToZip);

    expect($zipArchive->statIndex(0)['encryption_method'])->toBe(ZipArchive::EM_AES_256);

    $zipArchive->close();
});
```

pellaras:
thanks @freekmurze

PR #1983, opened 2026-08-20
Add support for --config option to override backup configuration

robert-abram:
Apply --config to the whole backup run, not just the command


freekmurze:
Thanks Robert! Nice catch that `--config` was only swapping the command's own config object while the zip layer and everything else resolving `Config` from the container kept using the default one. The two tests pinning down the password and relative path leaks are exactly what this needed.

I pushed a small follow-up removing the `backup-temporary-project` rebinding: the service provider binds it lazily off `config('backup.backup.temporary_directory')`, so your `config()->set('backup', ...)` above it already covers that case. Added a test for it too.

PR #1984, opened 2026-09-12
Fix an empty-string archive password silently enabling encryption

jeffersongoncalves:
## Summary

`env('BACKUP_ARCHIVE_PASSWORD')` returns `''` (not `null`) when the key is present but empty in `.env` — a very common setup for anyone who copies the published config and leaves the password blank to disable encryption, per the config's own comment (\"Set to `null` to disable encryption\").

`BackupConfig::fromArray()` only coalesces a *missing* key:

```php
password: $data['password'] ?? null,
```

An empty string is not `null`, so it passes straight through. `Zip::__construct()` / `Zip::open()` / `Zip::add()` treat any non-null password as "encryption wanted" and call `ZipArchive::setEncryptionName()` with that empty string. libzip then throws `ZipArchive::close(): Invalid argument` while finalizing the archive — with nothing in the error pointing at the password, since the failure surfaces generically at `close()` regardless of whether the zip contains 1 file or 20.

Traced this down in production by reproducing `addFile()` + `setCompressionName()` + `close()` in isolation against the real dump file (always succeeded), then finally checking `config('backup.backup.password')` directly and finding `string(0) ""` instead of `NULL`.

## Fix

Coerce a blank password to `null` in `BackupConfig::fromArray()`, so a config value like `'password' => env('BACKUP_ARCHIVE_PASSWORD')` actually disables encryption when the env var is left blank — matching the documented behavior.

## Test plan

- [x] Added `it treats an empty string password the same as a missing one` / `it keeps a real password` to `tests/Config/ConfigTest.php`
- [x] `composer analyse` (PHPStan) passes
- [x] `vendor/bin/pest --filter=ConfigTest` passes


freekmurze:
Thanks for tracking this down so thoroughly, @jeffersongoncalves! I made a small tweak so a whitespace-only password still enables encryption, and added an end to end test for the empty string. This is now released in [10.3.3](https://github.com/spatie/laravel-backup/releases/tag/10.3.3).

